PR #9, opened 2021-01-10
Add environment variable

leddest:


app/database/database.py
```diff
@@ -1,9 +1,14 @@
+import os
+
+from dotenv import load_dotenv
```
yammesicka, line 3: Let's use `os.getenv` instead of installing additional 3rd parties

app/database/database.py
```diff
@@ -1,9 +1,14 @@
...
 from sqlalchemy import create_engine
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker
 
 
-SQLALCHEMY_DATABASE_URL = "sqlite:///./dev.db"
+load_dotenv()
+
+SQLALCHEMY_DATABASE_URL = os.getenv("DB_URL")
```
yammesicka, line 11: The formal name is "database connection string".

Maybe we should change the name to `DATABASE_CONNECTION_STRING`?

yammesicka (changes requested):

PR #62, opened 2021-01-16
Feature/weather forecast

hadaskedar2020:
This feature gets weather data for given date & location.
It returns the following:
- Status - success / failure (+ ErrorDescription).
- Temprature data (min&max degrees) in Celsius &  Fahrenheit.
- ForecastType: forecast/ history/ historical average.
- Address - The location found by the service.


app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+        (str) - location / error description.
+    """
+    input_query_string = BASE_QUERY_STRING
+    input_query_string["startDateTime"] = input_date.isoformat()
+    input_query_string["endDateTime"] = (input_date + timedelta(days=1)).isoformat()
+    input_query_string["location"] = location
+    success_in_get_weather_data, api_json = get_data_from_api(HISTORY_URL, input_query_string)
+    if success_in_get_weather_data:
+        for item in api_json:
+            # print("historical:", api_json[item]['values'][0]['mint'], api_json[item]['values'][0]['maxt'])
```
Lisafiluz, line 93: necessary?

Lisafiluz (approved):
Looks great! :)

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+            output["ForecastType"] = forecast_type
+            output["Address"] = description
+    else:
+        output["Status"] = ERROR_STATUS
+        output["ErrorDescription"] = "Invalid date input provided"
+    return output
+
+
+if __name__ == "__main__":
+    print(get_weather_data("29", "02", 2024, "tel aviv"))
```
ivarshav, line 265: remove? :)

tests/test_weather_forecast.py
```diff
@@ -0,0 +1,30 @@
...
+    output = get_weather_data(day, month, year, location)
+    assert output['Status'] == expected
+
+
+def test_get_forecast_weather_data():
+    temp_date = datetime.now() + timedelta(days=1)
+    output = get_weather_data(temp_date.day, temp_date.month, temp_date.year, "tel aviv")
+    assert output['Status'] == 0
+
+# pytest.param(15, 1, 2021, "tel aviv", 0, id="basic forecast test"),
```
ivarshav, line 30: remove? :)

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
+from datetime import datetime, timedelta
```
ivarshav, line 1: I think this file should be in `internal\` folder

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+    'x-rapidapi-host': "visual-crossing-weather.p.rapidapi.com"
+    }
+BASE_QUERY_STRING = {"aggregateHours": "24", "unitGroup": "metric", "dayStartTime": "00:00:01",
+                     "contentType": "json", "dayEndTime": "23:59:59", "shortColumnNames": "True"}
+HISTORICAL_AVERAGE_NUM_OF_YEARS = 3
+OUTPUT = {"Status": SUCCESS_STATUS, "ErrorDescription": None, "MinTempCel": None, "MaxTempCel": None,
+          "MinTempFar": None, "MaxTempFar": None, "Conditions": None, "ForecastType": None}
+
+
+def validate_date_input(day, month, year):
```
ivarshav, line 26: Don't forget typing

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+        month (int) - month part of date.
+        year (int) - year part of date.
+    """
+    try:
+        day = int(day)
+        month = int(month)
+        year = int(year)
+    except ValueError:
+        return False, day, month, year
+    if 1975 <= year <= 2050:
```
ivarshav, line 44: magic numbers in code.. move to const?

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+    load_dotenv()
+    HEADERS['x-rapidapi-key'] = getenv('WEATHER_API_KEY')
+    success_in_get_weather_data = True
+    response = requests.request("GET", url, headers=HEADERS, params=input_query_string)
+    try:
+        response_json = response.json()["locations"]
+    except KeyError:
+        success_in_get_weather_data = False
+        response_json = response.json()
+    return success_in_get_weather_data, response_json
```
Gonzom, line 72: Check whether the request retuned a valid 200 response. There is also no need for a custom success boolean.
```
    response = requests.request("GET", url, headers=HEADERS, params=input_query_string)
    if response:
        try:
            return response.json()["locations"]
        except KeyError:
            return None, response.json()["message"]
    else:
        return None, "error message"   # TODO: valid error message
```

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+        (int) - maximum degrees in Celsius.
+        (str) - weather conditions.
+        (str) - location / error description.
+    """
+    input_query_string = BASE_QUERY_STRING
+    input_query_string["startDateTime"] = input_date.isoformat()
+    input_query_string["endDateTime"] = (input_date + timedelta(days=1)).isoformat()
+    input_query_string["location"] = location
+    success_in_get_weather_data, api_json = get_data_from_api(HISTORY_URL, input_query_string)
+    if success_in_get_weather_data:
```
Gonzom, line 91: After changing the above , then:
```
api_json, error_message = get_data_from_api(HISTORY_URL, input_query_string)
if api_json:
     ....
else:
     return None, None, None, error_message
```

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+        for item in api_json:
+            # print("historical:", api_json[item]['values'][0]['mint'], api_json[item]['values'][0]['maxt'])
+            min_temp = api_json[item]['values'][0]['mint']
+            max_temp = api_json[item]['values'][0]['maxt']
+            conditions = api_json[item]['values'][0]['conditions']
+        return min_temp, max_temp, conditions, api_json[item]['address']
```
Gonzom, line 97: From the indent of the return, the values of only the last instance are returned. Is that the intention? If so, then there isn't a need to iterate over the whole list.

Also remove commented out code.

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+        try:
+            datetime(year=year, month=month, day=day)
+        except ValueError:
+            return False, day, month, year
+    else:
+        return False, day, month, year
+    return True, day, month, year
+
+
+def get_data_from_api(url, input_query_string):
```
ivarshav, line 54: The name of the function is very general while it's goal is more specific..consider rename.

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+    if success_in_get_weather_data:
+        for item in api_json:
+            for i in range(len(api_json[item]['values'])):
+                if input_date == datetime.fromisoformat(api_json[item]['values'][i]['datetimeStr'][:-6]):
+                    min_temp = api_json[item]['values'][i]['mint']
+                    max_temp = api_json[item]['values'][i]['maxt']
+                    conditions = api_json[item]['values'][i]['conditions']
+        return min_temp, max_temp, conditions, api_json[item]['address']
+    else:
+        return None, None, None, api_json['message']
```
Gonzom, line 125: Same comments as above.

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+            "historical average" - average of the last 3 years on that date.
+                                    relevant for future dates (more then forecast).
+        min_temp (int) - minimum degrees in Celsius.
+        max_temp (int) - maximum degrees in Celsius.
+        conditions (str) - weather conditions.
+        Description (str) - location / error description.
+    """
+    input_date = datetime(year=year, month=month, day=day)
+    forecast_type = calculate_forecast_type(input_date)
+    if forecast_type == HISTORY_TYPE:
```
ivarshav, line 211: Maybe you can you a dict instead of many ifs?

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+    for relevant_year in relevant_years:
+        input_date = datetime(year=relevant_year, month=month, day=day)
+        min_temp, max_temp, conditions, description = get_historical_weather(input_date, location)
+        if min_temp is not None:
+            sum_min += min_temp
+            sum_max += max_temp
+    if min_temp is not None:
+        return sum_min / HISTORICAL_AVERAGE_NUM_OF_YEARS, sum_max / HISTORICAL_AVERAGE_NUM_OF_YEARS, description
+    else:
+        return None, None, description
```
Gonzom, line 170: Remove the extra parenthesis and place is directly in the for.
`for relevant_year in get_relevant_years_for_historical_average(day, month):`

If get_historical_weather(input_date, location) returns an error, such as a site down, bad api key, invalid address, then the function will still call the API x amount of times. Should it?

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+    input_date = datetime(year=year, month=month, day=day)
+    forecast_type = calculate_forecast_type(input_date)
+    if forecast_type == HISTORY_TYPE:
+        min_temp, max_temp, conditions, description = get_historical_weather(input_date, location)
+    if forecast_type == FORECAST_TYPE:
+        min_temp, max_temp, conditions, description = get_forecast_weather(input_date, location)
+    if forecast_type == HISTORICAL_AVERAGE_TYPE:
+        min_temp, max_temp, description = get_historical_average_weather(day, month, location)
+        conditions = ""
+    return forecast_type, min_temp, max_temp, conditions, description
```
Gonzom, line 218: You could simplify the code by:
```
    if forecast_type == HISTORY_TYPE:
        return forecast_type, get_historical_weather(input_date, location)
    if forecast_type == FORECAST_TYPE:
        return forecast_type, get_forecast_weather(input_date, location)
    if forecast_type == HISTORICAL_AVERAGE_TYPE:
        min_temp, max_temp, description = get_historical_average_weather(day, month, location)
        return forecast_type, min_temp, max_temp, "", description
```

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+        output_type (str) - "forecast" / "history" / "historical average".
+    """
+    delta = (input_date - datetime.now()).days
+    if delta < -1:
+        output_type = HISTORY_TYPE
+    elif delta > 15:
+        output_type = HISTORICAL_AVERAGE_TYPE
+    else:
+        output_type = FORECAST_TYPE
+    return output_type
```
Gonzom, line 187: The function returns a forecast type, so the name should be "get_forecast_type".

app/weather_forecast.py
Gonzom, line 265: Needed?

tests/test_weather_forecast.py
Gonzom, line 30: Remove commented out code.

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+        (list) - relevant years range.
+    """
+    if datetime.now() > datetime(year=datetime.now().year, month=month, day=day):
+        last_year = datetime.now().year
+    else:
+        last_year = datetime.now().year - 1
+    return list(range(last_year, last_year - HISTORICAL_AVERAGE_NUM_OF_YEARS, -1))
+
+
+def get_historical_average_weather(day, month, location):
```
yammesicka, line 144: Unfortunately, this function is not scalable.

We can't send 50 API requests from the server for 3 reasons:

1. The API key will be depleted very fast.
2. If the request is [synchronous](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests), it will cause the server to hang for a long time.
3. The load on the server gonna be huge for ~10k users.

There are two possible solutions: 

1. If there is an API service that supports this kind of historical querying, I suggest we use it.
2. If you can't find such an API service, we should (unfortunately) drop the historical average feature.

tests/test_weather_forecast.py
```diff
@@ -0,0 +1,30 @@
+from datetime import datetime, timedelta
```
yammesicka, line 1: Great job with the tests here :)

yammesicka (changes requested):

app/weather_forecast.py
hadaskedar2020: Could not find a free API that provides that.
So I've changed the logic to find the last history data for the relevant day & month (best forecast)
29.02 is supported by trying to check if the date is valid for the relevant year & only if not we will try to get data for the 28'th.

app/weather_forecast.py
```diff
@@ -0,0 +1,265 @@
...
+def get_data_from_api(url, input_query_string):
+    """ get the relevant weather data by calling the "Visual Crossing Weather" API.
+    Args:
+        url (str) - API url.
+        input_query_string (dict) - input for the API.
+    Returns:
+        success_in_get_weather_data (bool) - did the API call ended in success or failure (location not found etc).
+        response_json (json dict) - relevant part (data / error) of the JSON returned by the API.
+    """
+    load_dotenv()
```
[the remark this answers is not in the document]
hadaskedar2020: Isn't that the way to load env params from the file?

app/.env
```diff
@@ -0,0 +1,2 @@
+WEATHER_API_KEY=
```
yammesicka, line 1: Please move these to config.py.example. The default should be `os.getenv('WEATHER_API_KEY')`.

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,232 @@
...
+def get_data_from_weather_api(url, input_query_string):
+    """ get the relevant weather data by calling the "Visual Crossing Weather" API.
+    Args:
+        url (str) - API url.
+        input_query_string (dict) - input for the API.
+    Returns:
+        (json) - JSON data returned by the API.
+        (str) - error message.
+    """
+    load_dotenv()
```
yammesicka, line 53: We can eliminate this line (we work with `config.py`)

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,232 @@
...
+    HEADERS['x-rapidapi-key'] = os.getenv('WEATHER_API_KEY')
```
yammesicka, line 54: Here you can use config.WEATHER_API_KEY

requirements.txt
```diff
@@ -25,3 +25,6 @@ typing-extensions==3.7.4.3
 uvicorn==0.13.3
 wsproto==1.0.0
 zipp==3.4.0
+
+requests~=2.25.1
```
yammesicka, line 29: requirements.txt is usually kept sorted by alphabetic order.

(and, we can eliminate python-dotenv)

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,232 @@
...
+        location (str) - location name.
+    Returns:
+        weather_data (json) - output weather data.
+        error_text (str) - error message.
+    """
+    input_query_string = BASE_QUERY_STRING
+    input_query_string["location"] = location
+    api_json, error_text = get_data_from_weather_api(FORECAST_URL, input_query_string)
+    location_found = list(api_json.keys())[0]
+    if api_json:
```
yammesicka, line 106: It might be better to use 
```
if not api_json:
    return None
for i in range(...
```

In order to flatten the code a bit.

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,232 @@
...
+    if isinstance(requested_date, datetime.date):
+        if MIN_HISTORICAL_YEAR <= requested_date.year <= MAX_FUTURE_YEAR:
+            return True, None
+        else:
+            return False, INVALID_YEAR
+    else:
+        return False, INVALID_DATE_INPUT
+
+
+def get_data_from_weather_api(url, input_query_string):
```
yammesicka, line 44: We should cache it somehow to prevent long loading times.

Bonus: it might be better to use async/await here, if possible.

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,232 @@
...
+        (json) - JSON data returned by the API.
+        (str) - error message.
+    """
+    load_dotenv()
+    HEADERS['x-rapidapi-key'] = os.getenv('WEATHER_API_KEY')
+    try:
+        response = requests.request("GET", url, headers=HEADERS, params=input_query_string)
+    except requests.exceptions.RequestException:
+        return None, NO_API_RESPONSE
+    if response:
```
yammesicka, line 59: You can use if `response.is_ok` to be more specific here :)

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,232 @@
...
+    input_query_string["location"] = location
+    api_json, error_text = get_data_from_weather_api(HISTORY_URL, input_query_string)
+    location_found = list(api_json.keys())[0]
+    if api_json:
+        weather_data = {'MinTempCel': api_json[location_found]['values'][0]['mint'],
+                        'MaxTempCel': api_json[location_found]['values'][0]['maxt'],
+                        'Conditions': api_json[location_found]['values'][0]['conditions'],
+                        'Address': location_found}
+        return weather_data, None
+    else:
```
yammesicka, line 89: You can remove the `else` clause here :)

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,232 @@
...
+        error_text (str) - error message.
+    """
+    input_query_string = BASE_QUERY_STRING
+    input_query_string["startDateTime"] = input_date.isoformat()
+    input_query_string["endDateTime"] = (input_date + datetime.timedelta(days=1)).isoformat()
+    input_query_string["location"] = location
+    api_json, error_text = get_data_from_weather_api(HISTORY_URL, input_query_string)
+    location_found = list(api_json.keys())[0]
+    if api_json:
+        weather_data = {'MinTempCel': api_json[location_found]['values'][0]['mint'],
```
yammesicka, line 84: Can we be absolutely sure there is no `['values']` list with 0 items in it?

yammesicka (changes requested):

yammesicka:
Awesome progress, I've reviewed the code again and there are some small fixes left. GJ!

app/routers/weather_forecast.py
hadaskedar2020: In this phase we are after finding the "location" tag in the json, if the API had any problems we would have found them earlier. I don't think a try statement is required here.

app/routers/weather_forecast.py
hadaskedar2020: I've added cache support :-)

app/.env
hadaskedar2020: While trying to push my commits, I've encountered a problem: All checks have failed
I think it is probably because there is no API KEY for the tests. :-(
How will the actual KEY be set?
The status of the push is error due to that...

yammesicka: Read about API Mocking to prevent using the API (which might cost money, or have rate limit) every time we run the tests.

It will also solve this problem :)

hadaskedar2020: Thanks!

hadaskedar2020: Thanks!

tests/test_weather_forecast.py
```diff
@@ -0,0 +1,79 @@
+import datetime
```
yammesicka, line 1: Excellent job!! That's some advanced testing concepts you use here. Kudos!

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,272 @@
...
+    if isinstance(requested_date, datetime.date):
+        if MIN_HISTORICAL_YEAR <= requested_date.year <= MAX_FUTURE_YEAR:
+            return True, None
+        else:
+            return False, INVALID_YEAR
+
+
+def freezeargs(func):
+    """Transform mutable dictionary into immutable
+    Credit to 'fast_cen' from 'stackoverflow'
```
yammesicka, line 48: Can you please add a URL? It is needed in CC-BY-SA 4

app/routers/weather_forecast.py
```diff
@@ -0,0 +1,272 @@
+import datetime
```
yammesicka, line 1: That's should be in `app/internals/weather.py` (sorry if I said other thing in the past).

We don't want to give the users the option to use our API key for weather requests.

yammesicka (changes requested):

tests/test_weather_forecast.py
hadaskedar2020: Wow, blushing...

app/routers/weather_forecast.py
hadaskedar2020: I'm on it. Few min and I'll push it again. 10x.

PR #65, opened 2021-01-17
Style/grid

aviadamar:
**FRONT**
- Creating a layout.html page for inheritance in other pages.
- Creating a monthview.html: Calendar main layout for monthly view.
-  Creating CSS class for base web theme according to mockup. 

**BACK**
- Creating a temporary rout "monthview" for monthly view representation. 
- Creating a library calendar_grid.py - for all functions deals with managing calendar grid dates information.


app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
...
+
+DAYS_OF_THE_WEEK: List[str] = ['Monday', 'Tuesday', 'Wednesday',
+                               'Thursday', 'Friday', 'Saturday', 'Sunday']
+
+
+def get_date_as_string(date: datetime) -> str:
+    return date.strftime("%A %B %b %Y").split()
+
+
+def get_next_date(date):
```
yammesicka, line 13: Add type annotations

app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
...
+    """Generate date objects from a starting given date."""
+    while True:
```
yammesicka, line 15: Prefer `yield from (date + datetime.timedelta(days=i) for i in itertools.count(start=1))`

app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
...
+        date = next_day
+
+
+def get_day_n_before(date: datetime, n: int) -> datetime:
+    return date - datetime.timedelta(days=n)
+
+
+def get_n_days(date: datetime, n: int) -> List[datetime]:
+    g = get_next_date(date)
+    days = []
```
yammesicka, line 27: Prefer `yield from itertools.islice(g, days)`

app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
...
+    for day in range(n):
+        days.append(next(g))
+    return days
+
+
+def arrange_weeks(dates: List, n: int) -> List[List]:
```
yammesicka, line 33: Better type annotation required (what is the datatype inside the lists?)

app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
...
+    """Returns a list of lists with length of n for a given list"""
```
yammesicka, line 34: What is `n`?

app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
...
+    weeks = []
+    week = []
+    for day in dates:
+        if len(week) > 0 and len(week) % n - 1 == 0:
```
yammesicka, line 38: `if week and`...

app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
...
+            week = []
+        else:
+            week.append(day)
+    return weeks
+
+
+def get_month_block(date) -> List[List]:
+    days = calendar.monthrange(date.year, date.month)
+    before = get_day_n_before(date, days[0] + 1)
+    cal = get_n_days(before, 1000)
```
yammesicka, line 50: Should we really fetch 1000 days?

Anyhow, put magic numbers in constants with names.

app/main.py
```diff
@@ -18,15 +20,20 @@ def home(request: Request):
...
 
-@app.get("/profile")
-def profile(request: Request):
-
-    # Get relevant data from database
-    upcouming_events = range(5)
-    current_username = "Chuck Norris"
-
-    return templates.TemplateResponse("profile.html", {
+@app.get("/monthview")
```
yammesicka, line 23: Why have you deleted the `profile` function?

app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
+import calendar
```
yammesicka, line 1: Change the path.

Also, use `git rm` to remove the `app/List[List]` files.

yammesicka, line 1: This file should be on app/routing/dates.py, see https://fastapi.tiangolo.com/tutorial/bigger-applications/

yammesicka (changes requested):

yammesicka:
Where are your tests?

yammesicka:
Please change the PR to merge into PythonFreeCourse:develop

app/calendar_grid.py
```diff
@@ -0,0 +1,51 @@
...
+import datetime
+from typing import List, Tuple
+
+DAYS_OF_THE_WEEK: List[str] = ['Monday', 'Tuesday', 'Wednesday',
+                               'Thursday', 'Friday', 'Saturday', 'Sunday']
```
Gonzom, line 6: This both hardcodes the language and the order of the dates which it should not.
calendar.day_name - gives you a list of week names in the requested locale.
calendar.setfirstweekday(calendar.SUNDAY) - sets the week to start on a Sunday, which is how it starts in Israel.

Use these instead.

You can change the locale used for the names with locale.setlocale(locale.LC_TIME, ("fr")). 
Hebrew should work with locale.setlocale(locale.LC_TIME, ("he", "UTF-8")) but couldn't get that to correctly show Hebrew text.

aviadamar: True, its very important issue. 
For now I made it as a temporary var for displaying the overall layout,
There is an open ticket deals with this and its suppose to change it according the location. 
If it will not be done soon, maybe Ill do it by my self.

aviadamar: That was really helpful.

PR #70, opened 2021-01-17
feat: gets a link to the suitable image of a given event content.

nir9696:


app/event_images/event_images.py
```diff
@@ -0,0 +1,148 @@
+from nltk.tokenize import word_tokenize
```
ivarshav, line 1: Don't forget tests :)

app/event_images/event_images.py
```diff
@@ -0,0 +1,148 @@
...
+from typing import Optional
+from word_forms.lemmatizer import lemmatize
+
+import re
+
+
+IMAGES_LINK_MAP = {
```
ivarshav, line 8: I'd like to suggest an idea so you could discard this dict:
1. Rename all the images to be with exactly the same name as the key (for ex.: cycling.jpg -> cycle.jpg).
2. Create an auxiliary function which receives the key and generates the full path: f"event_flairs\{key}.jpg".

Also, note that you are using "\".
It's better to verify on which os the application is running on.
Even better - support all os platforms :)

app/event_images/tests/test_event_images.py
```diff
@@ -0,0 +1,60 @@
+from event_images import attach_image_to_event, get_key, IMAGES_LINK_MAP, IMAGES_RELATED_WORDS_MAP, remove_non_alphabet_chars, search_token_in_related_words
```
ivarshav, line 1: Please reformat the file (too many blank lines between pieces of code).

app/event_images/event_images.py
nir9696: good idea! I'll do that!

app/event_images/event_images.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Args:
+        event_content (str): The event content.
+
+    Returns:
+        str: The link to the suitable image of a given token content.
+    """
+    event_tokens = word_tokenize(event_content)
+    for token in event_tokens:
+        if re.match(r'\w', token):
```
yammesicka, line 122: `if token[0].isalnum()`?

app/event_images/event_images.py
```diff
@@ -0,0 +1,133 @@
...
+        str: The value's key in IMAGES_RELATED_WORDS_MAP dictionary..
+    """
+    shrunken = remove_non_alphabet_chars(val).lower()
+    for key, values in IMAGES_RELATED_WORDS_MAP.items():
+        shrunken_list = [remove_non_alphabet_chars(v).lower() for v in values]
+        if shrunken in shrunken_list:
+            return key
+
+
+def search_token_in_related_words(token: str) -> Optional[str]:
```
yammesicka, line 97: We might want to cache this function

yammesicka (changes requested):

yammesicka:
Please commit to PythonFreeCourse:develop instead

app/event_images/tests/test_event_images.py
```diff
@@ -0,0 +1,73 @@
+from event_images import attach_image_to_event, get_key,\
```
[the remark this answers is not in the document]
nir9696: I know :) I broke the line because it was too long according to flake8...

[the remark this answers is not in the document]
ivarshav: I think he meant the path : `app/event_images/tests/test_event_images.py`

app/event_images/event_images.py
nir9696: can you hint me how do I support all os platforms?

app/event_images/tests/test_event_images.py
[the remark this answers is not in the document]
nir9696: Done

app/event_images/event_images.py
yammesicka: I'd go with something bigger. 4,096 words will include about 95% of the usages in English, and this is pretty reasonable space to allocate in memory :)

app/event_images/event_images.py
```diff
@@ -0,0 +1,216 @@
+from functools import lru_cache
```
yammesicka, line 1: Can you please move this file to app/routers/event_images.py?

The images should be in app/static/event_images/...

yammesicka (changes requested):

yammesicka:
Great job! I've added 2 more comments to fix before the merge.

app/event_images/event_images.py
nir9696: done :) these are the 2 comments you mentioned?

tests/test_event_images.py
```diff
@@ -0,0 +1,73 @@
...
+    assert search_token_in_related_words(token) == link
+
+
+event_contents = [
+    (r"memo backrub and medicines!!!!", r'..\static\event_flairs\massage.jpg'),
+    (r"Dont forget medicines & backrub!!", r'..\static\event_flairs\pill.jpg'),
+    (r"It's important to drink", r'..\static\event_flairs\drank.jpg'),
+    (r"call Jim about tennis on friday", r'..\static\event_flairs\tennis.jpg'),
+    (r"have to check on pikachu", r'#'),
+    (r"-~                   new pokemon episode 19:00 ~!", r'..\static\event_flairs\pokemon.jpg'),
```
PureDreamer, line 67: The lines are too long, the build won't pass, please fix :)

app/routers/event_images.py
```diff
@@ -0,0 +1,216 @@
...
+        str: The suitable link.
+    """
+    return f'{FLAIRS_REL_PATH}\\{lemmatized_word}{FLAIRS_EXTENSION}'
+
+
+def remove_non_alphabet_chars(s: str) -> str:
+    """Remove non-alphabet chars from a given string
+
+    Args:
+        s (str): The string to remove the non-alphabet chars from.
```
PureDreamer, line 157: Would be better to change the name to something recognizable like text, word etc'

app/routers/event_images.py
```diff
@@ -0,0 +1,216 @@
+from functools import lru_cache
+from nltk.tokenize import word_tokenize
+from typing import Optional
+from word_forms.lemmatizer import lemmatize
+
+import re
+
+
+FLAIRS_EXTENSION = '.jpg'
+FLAIRS_REL_PATH = r'..\static\event_flairs'
```
yammesicka, line 10: Awesome progress, nearly done in my opinion.

It might be a bit dangerous to use `..\static` in your code.
If we change the project structure, the code will break and the tests won't find it.

Can you somehow make the `static`  directory configurable?
Maybe we should put it in `config` and make it absolute using `os.path.absolute`

yammesicka (changes requested):

app/event_images/event_images.py
ivarshav: Try to google it :)

app/routers/event_images.py
nir9696: Done!

yammesicka:
Unfortunately I had to truncate your changes to `requirements.txt`, as your commit added lots of unused packages (such as JupyterNotebook).

Please add only the relevant packages to the requirement file and push again.

app/config.py
```diff
@@ -0,0 +1,4 @@
+import os
+
+
+STATIC_ABS_PATH = os.path.abspath("static")
```
yammesicka, line 4: This should be in app/config.py.example

(Please `git rm` the `app/config.py` file)

yammesicka: Great job! Just fix this one and we're good to go.

yammesicka (changes requested):

nir9696:
In the linter comments by flake8, I've gor "AttributeError: module 'app.config' has no attribute 'STATIC_ABS_PATH'",
but it has! I spent many many hours to figure out what the problem is and I failed to understand this.
I saw many questions about import errors like mine in stackoverflow, with no satisfying answers... I've also tried to ask my peers in the course but I didn't get any response.
can you help me?

yammesicka:
You could ask in the forums :)

Our CI copy the `config.py.example` and overrides any instance of `config.py`.

You should put the configurations in `config.py.examples`.

app/config.py
nir9696: done

app/config.py.example
```diff
@@ -26,10 +26,13 @@ PRODUCT_ID = '-//Our product id//'
 email_conf = ConnectionConfig(
     MAIL_USERNAME=os.getenv("MAIL_USERNAME") or "user",
     MAIL_PASSWORD=os.getenv("MAIL_PASSWORD") or "password",
-    MAIL_FROM=os.getenv("MAIL_FROM") or "a@a.com",
+    MAIL_FROM=os.getenv("MAIL_FROM") or "a@a.com",
```
yammesicka, line 29: Please remove the extra whitespaces added after this line

yammesicka (approved):

yammesicka:
Great job, but pytest is failing during the CI. Can you please fix it?

PR #83, opened 2021-01-18
feat: Add a feature for sharing information with a WhatsApp account

nelliei:
This feature can be used by anyone to share information with a WhatsApp account.
I am working now on a basic feature that can demonstrate how it can be used by anyone.


app/internal/whatsapp.py
```diff
@@ -0,0 +1,23 @@
...
+    It takes a string message and a cell phone number and it sends the message to that phone number.
+    Args:
+        phone_number (str): Cell phone number to send the message to.
+        message (str): Message that is going to be sent.
+
+    Returns:
+        bool: Returns True if the message was sent, else returns False.
+    """
+    try:
+        webbrowser.open_new(f'https://api.whatsapp.com/send?phone={phone_number}&text={message}')
```
yammesicka, line 15: We can't open webbrowser, because the code isn't really running on the users' end.
We should give him a link or something instead :)

tests/test_whatsapp.py
```diff
@@ -0,0 +1,28 @@
...
+    ('972536106106', ''),
+    # You will redirect to whatsapp window of the specified number and you can write your own message
+    ('', 'Which phone number?'),
+    # You will redirect to whatsapp window and you can choose someone from your contact list
+]
+
+
+@pytest.mark.parametrize('phone_number, message', no_phone_or_msg)
+def test_no_message_or_phone(phone_number, message):
+    assert whatsapp.send(phone_number, message) == False
```
yammesicka, line 28: `assert not whatsapp.send(...`

tests/test_whatsapp.py
```diff
@@ -0,0 +1,28 @@
...
+    ('972536106106', 'Event or a joke or the schedule of one day'),
+    # You will redirect to whatsapp window of the specified number and you can send the message
+    ('999999', 'Wrong phone number?'),
+    # You will redirect to whatsapp window and you will receive a message "The phone number shared via a link is incorrect"
+]
+
+
+@pytest.mark.parametrize('phone_number, message', phone_and_msg)
+def test_whatsapp_send(phone_number, message):
+    assert whatsapp.send(phone_number, message) == True
```
yammesicka, line 15: `== True` is redundant.

yammesicka (changes requested):

app/internal/whatsapp.py
nelliei: I have changed the function, and now it returns a link :)

tests/test_whatsapp.py
nelliei: Thank you! I have changed the tests according to the new method

app/internal/whatslink.py
```diff
@@ -0,0 +1,18 @@
+from typing import Optional
```
yammesicka, line 1: I think it would be easier to understand if the name of the file would be "whatsapp" :)

app/internal/whatslink.py
```diff
@@ -0,0 +1,18 @@
...
+    can add it to an html page and send the message to that phone number.
+    Args:
+        phone_number (str): Cell phone number to send the message to.
+        message (str): Message that is going to be sent.
+
+    Returns:
+        str: Returns a string which contains a link to whatsapp api so we can
+             send the message via whatsapp.
+    """
+    message = message.replace(" ", "%20")
```
yammesicka, line 16: Maybe we should `urlencode` the message, 'cause there are more special characters (not only spaces)

yammesicka (changes requested):

yammesicka:
Good work!

I'll probably decide to leave this PR opened until we'll have a way to integrate it into the frontend :)

app/internal/whatsapp.py
```diff
@@ -0,0 +1,20 @@
...
+    Args:
+        phone_number (str): Cell phone number to send the message to.
+        message (str): Message that is going to be sent.
+
+    Returns:
+        str: Returns a string which contains a link to whatsapp api so we can
+             send the message via whatsapp.
+    """
+    link = 'https://api.whatsapp.com/send?'
+    mydict = {'phone': f'{phone_number}', 'text': f'{message}'}
```
IdanPelled, line 18: why are you using f string here?

tests/test_whatsapp.py
```diff
@@ -0,0 +1,35 @@
...
+    assert whatsapp.make_link(phone_number, message) == 'https://'\
+        'api.whatsapp.com/send?phone=972536106106&text='
+
+
+def test_no_number():
+    # Redirects to whatsapp window. Choose someone from your own contact list.
+    phone_number = ""
+    message = 'Which phone number?'
+    assert whatsapp.make_link(phone_number, message) == 'https://'\
+        'api.whatsapp.com/send?phone=&text=Which+phone+number%3F'
```
IdanPelled, line 35: great work!

app/internal/whatsapp.py
nelliei: Thanks for the lesson :)   I have changed it for the next push :)

app/internal/whatsapp.py
```diff
@@ -0,0 +1,20 @@
+from typing import Optional
+from urllib.parse import urlencode
+
+
+def make_link(phone_number: Optional[str], message: Optional[str]) -> str:
```
yammesicka, line 5: This should have a route, like `/event/share/whatsapp` :)

I want to see this function in the [API](https://calendar.pythonic.guru/docs).
It will be automatically added when you add a route using FastAPI.

yammesicka (changes requested):

tests/test_whatsapp.py
```diff
@@ -0,0 +1,37 @@
+from app.routers import whatsapp
```
yammesicka, line 1: Please add an end to end test:

Use the client to fetch the `/whatsapp` route and check you get a valid JSON.

yammesicka (changes requested):

tests/test_whatsapp.py
```diff
@@ -0,0 +1,45 @@
...
+    message = 'Which phone number?'
+    assert whatsapp.make_link(phone_number, message) == {
+        "link": "https://api.whatsapp.com/send?phone=&text=Which+phone+"
+        "number%3F"}
+
+
+def test_end_to_end_testing(client):
+    resp = client.get('/whatsapp?phone_number=972536106106&message=testing')
+    assert resp.status_code == status.HTTP_200_OK
+    assert resp.content == b'{"link":"https://api.whatsapp.com/send?phone='\
```
yammesicka, line 44: Prefer using `.json`, to guarantee success even if the order of the elements is not the same.

tests/test_whatsapp.py
```diff
@@ -0,0 +1,45 @@
...
+    phone_number = ""
+    message = 'Which phone number?'
+    assert whatsapp.make_link(phone_number, message) == {
+        "link": "https://api.whatsapp.com/send?phone=&text=Which+phone+"
+        "number%3F"}
+
+
+def test_end_to_end_testing(client):
+    resp = client.get('/whatsapp?phone_number=972536106106&message=testing')
+    assert resp.status_code == status.HTTP_200_OK
```
yammesicka, line 43: Prefer `status_code.ok`, to guarantee success even if the resulted page refers to another page with status code 200.

yammesicka (changes requested):

PR #88, opened 2021-01-19
Feature/send email

annashtirberg:
Added a feature to send emails to invite others to use our app
this pull request includes:
route: to send email
internal code to send email
tests: to test the route and internal code
configuration: to set parameters for the SMTP server,
and can be reused in other places


app/config.py
```diff
@@ -0,0 +1,16 @@
+from pydantic import BaseSettings
```
ShiZinDle, line 1: This file seems to be feature specific and should be placed in a subpackage along with all other related files.

app/config.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Settings(BaseSettings):
```
ShiZinDle, line 4: The naming scheme here is not very indicative.
Consider more verbose names, especially if not placed in a subpackage

app/internal/mail.py
```diff
@@ -0,0 +1,65 @@
+import os
```
ShiZinDle, line 1: Can't see why this shouldn't be in the same subpackage as the rest of the feature's code.

app/internal/mail.py
```diff
@@ -0,0 +1,65 @@
...
+from pydantic.errors import EmailError
+
+import config
+
+templates = Jinja2Templates(directory=os.path.join("app", "templates"))
+
+# application name
+CALENDAR_SITE_NAME = "Calendar"
+# link to the home page of the application
+CALENDAR_HOME_PAGE = "calendar.com"
```
ShiZinDle, line 15: The correct path is 'calendar.pythonic.guru', nevertheless I reckon app.url_path_for()

app/internal/mail.py
```diff
@@ -0,0 +1,65 @@
...
+# link to the application registration page
+CALENDAR_REGISTRATION_PAGE = r"calendar\registration.com"
```
ShiZinDle, line 17: The top-level domain (.com) should be placed right after the host (i.e. calendar.com/registration).
consider app.url_oath_for() here as well.

app/routers/mail.py
```diff
@@ -0,0 +1,38 @@
...
+router = APIRouter()
+
+CORRECT_EMAIL_ADDRESS_ERROR_MESSAGE = "Please enter valid email address"
+SUCCESSFULLY_SENT_EMAIL_MESSAGE = "Your message was sent successfully"
+
+
+class InvitationParams(BaseModel):
+    sender_name: str
+    recipient_name: str
+    recipient_mail: str
```
ShiZinDle, line 17: Why isn't this in the config file?

app/routers/mail.py
```diff
@@ -0,0 +1,38 @@
+from fastapi import BackgroundTasks, APIRouter, Depends
+from pydantic import BaseModel, EmailStr
+from pydantic.errors import EmailError
+
+import config
+from internal.mail import send_fast_email_invitation
+
+router = APIRouter()
+
+CORRECT_EMAIL_ADDRESS_ERROR_MESSAGE = "Please enter valid email address"
```
ShiZinDle, line 10: Should be renamed to 'incorrect' or 'invalid' rather than 'correct'

conftest.py
```diff
@@ -0,0 +1,4 @@
+import os
+import sys
+
+sys.path.append(os.path.join(os.curdir, "app"))
```
ShiZinDle, line 4: This is considered a bad practice, and shouldn't be relevant if run from a different directory.

tests/test_mail.py
```diff
@@ -0,0 +1,164 @@
...
+            {'detail': [
+                {'loc': ['body', 'sender_name'],
+                 'msg': 'field required',
+                 'type': 'value_error.missing'},
+                {'loc': ['body', 'recipient_name'],
+                 'msg': 'field required',
+                 'type': 'value_error.missing'}
+            ]}
+    ),
+])
```
ShiZinDle, line 103: Consider placing in a variable for readability.

ShiZinDle (changes requested):

app/config.py
annashtirberg: I wanted to use the general configuration.
but there still isn't any, so this should be unioned with it later

app/config.py
annashtirberg: this is meant to be the general configuration for all of the application.
currently only email fields are relevant.
smtp is hinting that those are email related configurations

app/internal/mail.py
annashtirberg: in a different pull request I saw that Yam told him to split the route related code and the complicated stuff to seperate files
so this is what I did here

app/internal/mail.py
annashtirberg: I will change the name, thanks.
but the app.url_path_for will not work for two reasons:
look here: https://stackoverflow.com/questions/63682956/fastapi-retrieve-url-from-view-name-route-name
the results will be "/" or "/registration" which will not be good inside an email.
the other reason is that the internal part of the code doesn't use the app, or at least that how I used it

I will try to think of a better way to do this

app/internal/mail.py
annashtirberg: ok, I will fix the top-level domain. thanks

regarding the rest see above

app/routers/mail.py
annashtirberg: those are the parameters for the body of this route. they are not configurations

as you can see here [https://fastapi.tiangolo.com/tutorial/body/#create-your-data-model](BaseModel)

app/routers/mail.py
annashtirberg: I will corrent this. thank you

conftest.py
annashtirberg: ok, I didn't now this. I will fix this.
good to learn, thanks.

tests/test_mail.py
annashtirberg: I wrote a helper function to read the error responses
now the parameters are more readable

tests/test_mail.py
```diff
@@ -0,0 +1,163 @@
...
+        loc = detail["loc"]
+        assert isinstance(loc, list)
+        assert 2 == len(loc)
+        assert "body" == loc[0]
+        assert loc[1] in missing_fields
+
+
+def test_read_main():
+    response = client.get("/")
+    assert response.status_code == 200
```
yammesicka, line 62: Prefer `response.status_code.is_ok`.

#### Why?
The response code might be something like 302 - which means the user is redirected to another page.
If eventually the user lands on a page with the status code `200 OK`, the `is_ok` attribute will be `True`.

tests/test_mail.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def test_send_mail_no_body(configured_smtpd):
+    response = client.post("/mail/invitation/")
+    assert response.status_code == 422
```
yammesicka, line 67: Prefer `status.HTTP_422_UNPROCESSABLE_ENTITY` (read more about it [here](https://fastapi.tiangolo.com/tutorial/handling-errors/))

tests/test_mail.py
```diff
@@ -0,0 +1,163 @@
...
+    assert response.json() == {'detail': [{
+        'loc': ['body'],
+        'msg': 'field required',
+        'type': 'value_error.missing'}]}
+    assert len(configured_smtpd.messages) == 0
```
yammesicka, line 72: Prefer the Pythonic way: `assert not configured_smtpd.messages`

conftest.py
```diff
@@ -0,0 +1,6 @@
+import os
+import sys
+
+this_dir = os.path.dirname(__file__)
```
yammesicka, line 4: These three lines aren't needed :)

app/config.py
```diff
@@ -0,0 +1,16 @@
...
+    smtp_from_email: str = "invite@calendar-app.com"
+    smtp_port: int = 3999
+    smtp_server: str = "localhost"
+    smtp_use_tls: bool = False
+    smtp_use_ssl: bool = False
+    smtp_use_credentials: bool = False
+
+
+def get_settings():
+    return Settings()
```
yammesicka, line 16: That's a good way of doing so, but we already have a configuration file :) Please take a look at [`config.py.example`](https://github.com/PythonFreeCourse/calendar/blob/develop/app/config.py.example).

When the user clones the repo, he should copy it and either:

1. Create his own `config.py` with his own configurations
2. Create his own `config.py` and set his environment variables according to the variable names in the file.

app/main.py
```diff
@@ -1,13 +1,20 @@
...
 from fastapi.staticfiles import StaticFiles
 from fastapi.templating import Jinja2Templates
 
+from routers import mail
 
 app = FastAPI()
 
-app.mount("/static", StaticFiles(directory="static"), name="static")
+app.mount("/static",
+          StaticFiles(directory=os.path.join("app", "static")),
```
yammesicka, line 12: You sure this is necessary? :)

yammesicka (changes requested):
Great PR. I've added few of my insights, please take a look :)

conftest.py
annashtirberg: but then the tests fail because they import fails
Is all the file not needed?

app/config.py
annashtirberg: oyy
I need this for my mail tests because i use other configurations for tests then not tests
this way you can put environment vairables or default values
like in here: [https://fastapi.tiangolo.com/advanced/settings/#create-the-settings-object](url) 
and then i use this: [https://fastapi.tiangolo.com/advanced/settings/#settings-and-testing](url) for tests

in the mail team we worked a lot on making this work
is it ok if I add database and media sopport to this way of configuration?
all their tests work good for me when doing this

tests/test_mail.py
annashtirberg: ok, i use response.ok instead

tests/test_mail.py
annashtirberg: ok :)

tests/test_mail.py
annashtirberg: ok :)

conftest.py
annashtirberg: ok did it :)

app/main.py
annashtirberg: oyy no. fixed it

yammesicka:
Great job :) Please send the PR to :develop and not to :main, so I'll be able to clearly see the changes

app/config.py
annashtirberg: ok

app/internal/mail.py
annashtirberg: ok

app/config.py
annashtirberg: ok

tests/test_event_images.py
```diff
@@ -1,9 +1,9 @@
-from app.routers.event_images import attach_image_to_event,\
-    generate_flare_link_from_lemmatized_word, get_image_name,\
-    remove_non_alphabet_chars, search_token_in_related_words
-from app import config
 import pytest
 
+from app import config
+from app.routers.event_images import attach_image_to_event, \
```
yammesicka, line 4: Prefer splitting lines using parenthesis :)

app/internal/email.py
```diff
@@ -1,11 +1,27 @@
...
-from fastapi import BackgroundTasks
+from fastapi import BackgroundTasks, UploadFile
 from fastapi_mail import FastMail, MessageSchema
+from pydantic import EmailStr
+from pydantic.errors import EmailError
 from sqlalchemy.orm.session import Session
 
 mail = FastMail(email_conf)
 
+templates = Jinja2Templates(directory=os.path.join("app", "templates"))
```
yammesicka, line 16: Can you please move this, and the other 3 constant variables after this line, to the config file?

yammesicka (commented):
Awesome job! Sorry for the long waiting time. Added few really small, probably final remarks :)

app/internal/email.py
```diff
@@ -31,10 +47,136 @@ def send(
...
+    :param background_tasks: (BackgroundTasks): Function from fastapi that lets
+            you apply tasks in the background.
+    :return: bool, True if the file was successfully
+    sent to the recipient, and False if the entered
+    email address is incorrect or file does not exist.
+    """
+    if not verify_email_pattern(recipient_mail):
+        return False
+
+    if os.path.exists(file_path) is False:
```
yammesicka, line 124: Prefer: `if not os.path.exists(file_path)`

tests/test_event_images.py
yammesicka, line 4: Prefer splitting using parenthesis

yammesicka (changes requested):
Awesome job! Sorry for the long waiting time. Added few really small, probably final remarks :)

app/internal/email.py
annashtirberg: ok :)

tests/test_event_images.py
annashtirberg: ok :)

annashtirberg: ok :)

app/internal/email.py
annashtirberg: ok :)

yammesicka:
In `app/routers/email.py`, please add tests to:

```
raise HTTPException(status_code=422, detail="Couldn't send the email!")
```

In `app/internal/email.py`, please add coverage to `send_email_file` and `send_email_invitation`

yammesicka (approved):

Issue #35, opened 2021-01-12
Add an option of update event


efratush:
Allow the event creator(only) to update the event.
An email should be sent to all recipients of the event, informing them that the event has been updated.

PR #100, opened 2021-01-20
Add basic update event

efratush:
#35


app/routers/event.py
```diff
@@ -19,3 +24,73 @@ async def eventedit(request: Request):
...
+def get_event_by_id(db: Session, event_id: int) -> Event:
+    """Select event by id"""
+    return db.query(Event).filter(Event.id == event_id).first()
+
+
+# copy from michael ben david
+def check_validation(start_time: datetime, end_time: datetime) -> bool:
+    """Check if the start time is earlier than the end time"""
+
+    if start_time < end_time:
```
yammesicka, line 38: Prefer: `return start_time < end_time`.

app/routers/event.py
```diff
@@ -19,3 +24,73 @@ async def eventedit(request: Request):
...
                                       {"request": request, "event_id": id})
+
+
+def get_event_by_id(db: Session, event_id: int) -> Event:
+    """Select event by id"""
+    return db.query(Event).filter(Event.id == event_id).first()
+
+
+# copy from michael ben david
+def check_validation(start_time: datetime, end_time: datetime) -> bool:
```
yammesicka, line 35: Prefer a informative naming: `is_later_date`

app/routers/event.py
```diff
@@ -19,3 +24,73 @@ async def eventedit(request: Request):
...
+    """Check if the start time is earlier than the end time"""
+
+    if start_time < end_time:
+        return True
+    return False
+
+
+def get_column_names_event(db: Session) -> Set:
+    result = db.execute("select * from events")
+    return {col for col in result.keys()}
```
yammesicka, line 45: Prefer using `set()`

(Great job choosing `set`. btw)

app/routers/event.py
```diff
@@ -19,3 +24,73 @@ async def eventedit(request: Request):
...
+def check_validation(start_time: datetime, end_time: datetime) -> bool:
+    """Check if the start time is earlier than the end time"""
+
+    if start_time < end_time:
+        return True
+    return False
+
+
+def get_column_names_event(db: Session) -> Set:
+    result = db.execute("select * from events")
```
yammesicka, line 44: Prefer using the ORM instead.

Also, If possible, try to get the column names without selecting the whole table, which might be a very heavy query.

app/routers/event.py
```diff
@@ -19,3 +24,73 @@ async def eventedit(request: Request):
...
+                if not check_validation(event_update.start, old_event.end):
+                    return False
+            elif not check_validation(old_event.start, event_update.end):
+                return False
+        except TypeError:
+            return False
+    return True
+
+
+def update_event(event_id: int, event_items: Dict, db: Session
```
yammesicka, line 65: This function is really long. Can we somehow split it into multiple functions?

app/routers/event.py
```diff
@@ -19,3 +24,73 @@ async def eventedit(request: Request):
...
+    try:
+        old_event = get_event_by_id(db=db, event_id=event_id)
+    except AttributeError:  # Problem connecting to db
+        return None
+
+    if old_event is None:   # No such event number.
+        return None
+
+    column_names = get_column_names_event(db)
+    for key in event_items.keys():
```
yammesicka, line 82: You can leave out the `.keys`

app/routers/event.py
```diff
@@ -19,3 +24,73 @@ async def eventedit(request: Request):
...
+        if key not in column_names:
```
yammesicka, line 83: Another nice, readable trick might be:

```
if any(key not in column_names for key in event_items):
    return None
```

yammesicka (changes requested):

tests/test_event.py
```diff
@@ -1,22 +1,58 @@
...
+def test_eventedit(client):
     response = client.get("/event/edit")
     assert response.status_code == 200
     assert b"Edit Event" in response.content
 
 
-def test_eventview_with_id():
+def test_eventview_with_id(client):
     response = client.get("/event/view/1")
     assert response.status_code == 200
```
ivarshav, line 23: Please use starlette.status for status code

tests/test_event.py
```diff
@@ -1,22 +1,58 @@
...
 
-def test_eventview_without_id():
+def test_eventview_without_id(client):
     response = client.get("/event/view")
     assert response.status_code == 404
+
+
+@pytest.mark.parametrize("data", INVALID_UPDATE_OPTIONS)
+def test_invalid_update(data, event, session):
+    assert update_event(1, data, session) is None
```
ivarshav, line 34: What is 1? the event id?

efratush: fixed

tests/test_event.py
```diff
@@ -17,5 +28,36 @@ def test_eventview_without_id(self, client):
...
+                            event_dict=data, db=session) is None
+
+    @staticmethod
+    def test_successful_update(event, session):
+        data = {
+            "title": "successful",
+            "start": datetime(2021, 1, 20),
+            "end": datetime(2021, 1, 21),
+        }
+        assert type(update_event(1, data, session)) == Event
```
yammesicka, line 44: Use `isinstance` instead :)

app/routers/event.py
```diff
@@ -27,6 +28,41 @@ async def eventview(request: Request, id: int):
...
+    """Select event by id"""
+    return db.query(Event).filter(Event.id == event_id).first()
+
+
+def validate_dates(start_date: datetime, end_date: datetime) -> bool:
+    """Check if the start date is earlier than the end date"""
+    return start_date < end_date
+
+
+def update_event(event_id: int, event_dict: Dict, db: Session
```
yammesicka, line 41: Maybe just `events`? We usually try not to use [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation) when it's possible :)

app/routers/event.py
```diff
@@ -27,6 +28,41 @@ async def eventview(request: Request, id: int):
                                       {"request": request, "event_id": id})
 
 
+def get_event_by_id(db: Session, event_id: int) -> Event:
```
yammesicka, line 31: I'd event call it `by_id` to allow external modules use `event.by_id`

yammesicka (commented):
Great job! Added few of my insights :)

app/routers/event.py
efratush: *event

app/routers/event.py
```diff
@@ -27,6 +28,41 @@ async def eventview(request: Request, id: int):
                                       {"request": request, "event_id": id})
 
 
+def by_id(db: Session, event_id: int) -> Event:
+    """Select event by id"""
+    return db.query(Event).filter(Event.id == event_id).first()
+
+
+def validate_dates(start_date: datetime, end_date: datetime) -> bool:
```
yammesicka, line 36: Maybe `is_date_before`?

app/routers/event.py
```diff
@@ -27,6 +28,41 @@ async def eventview(request: Request, id: int):
...
+    """Check if the start date is earlier than the end date"""
+    return start_date < end_date
+
+
+def update_event(event_id: int, event: Dict, db: Session
```
yammesicka, line 41: This function does more than one thing. Can you please separate it to 2-3 function?

yammesicka (changes requested):

app/routers/event.py
```diff
@@ -27,6 +28,41 @@ async def eventview(request: Request, id: int):
...
+                 ) -> Optional[Event]:
+    # TODO Check if the user is the owner of the event.
+
+    # Extract only that keys to update
+    event_to_update = {i: event[i] for i in (
+        'title', 'start', 'end', 'content', 'location') if i in event}
+    if not bool(event_to_update):  # Event items is empty
```
yammesicka, line 48: Is the `bool` here necessary?

yammesicka (changes requested):

app/routers/event.py
efratush: You're right.

yammesicka (approved):

PR #125, opened 2021-01-23
WIP Speedy Meetings

omrizhaki:
* Created a setting in the DB that will record whether a user has Speedy Meetings enabled
* Added an option to use default_end time, in case you don't want the user to decide when the meeting will end
* Added a function that will shorten the default meeting time in case Speedy Meetings is enabled


app/database/models.py
```diff
@@ -29,6 +34,24 @@ class Event(Base):
     content = Column(String)
     start = Column(DateTime, nullable=False)
     end = Column(DateTime, nullable=False)
+    default_end = get_default_end_time()
```
yammesicka, line 37: This class is part of the ORM, it must be of type `Column` or a `relationship`.

You can check the SQLAlchemy docs to see how to add default value to the `end` column.

yammesicka (changes requested):

app/database/models.py
```diff
@@ -29,6 +34,24 @@ class Event(Base):
...
     owner_id = Column(Integer, ForeignKey("users.id"))
 
     owner = relationship("User", back_populates="events")
+
+    user = get_user_by_id(owner_id)
+
+    def get_user_by_id(self, owner_id):
```
ivarshav, line 44: Is it needed? can't we just call user.owner?

app/database/models.py
```diff
@@ -29,6 +34,24 @@ class Event(Base):
...
+        return owner
+
+    def get_default_end_time(self):
+        return start + get_event_duration()
+
+    def get_event_duration(self):
+        default_duration = get_default_duration()
+        if user.has_speedy_meetings_enabled():
+            return default_duration * .75
```
ivarshav, line 53: what is 0.75?

app/database/models.py
```diff
@@ -29,6 +34,24 @@ class Event(Base):
...
+        return default_duration
+
+    def get_default_duration(self):
```
ivarshav, line 56: Why it's part of the table model?

PR #128, opened 2021-01-24
feat: add a daily inspirational quote

zohary89:
- load quotes from a json file (copied from a free api) to the db 
- the daily quote is displayed in the home page, until the calendar view will be developed.
- add tests
* extra small change in agenda.html- I add a link  from the event's title to the event's view page.


app/database/models.py
```diff
@@ -74,3 +74,11 @@ def __repr__(self):
             f'({self.event.owner}'
             f'to {self.recipient})>'
         )
+
+
+class Quote(Base):
+    __tablename__ = "quotes"
+
+    id = Column(Integer, primary_key=True, index=True)
+    text = Column(String)
```
ivarshav, line 83: Can this property be null?

app/internal/quotes/daily_quotes.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def quote_per_day(
+        session: Session, date: date = date.today()
+        ) -> Optional[Quote]:
+    """This function provides a daily quote, relevant to the current
+    day of the year. The quote is randomally selected from a set
+    of quotes matching to the given day"""
+    day_num = date.timetuple().tm_yday
+    quotes = session.query(Quote).filter(Quote.id % 366 == day_num).all()
```
ivarshav, line 17: Consider const for the number

app/main.py
```diff
@@ -13,16 +15,22 @@
...
+load_quotes.load_daily_quotes(next(get_db()))
+
 app.include_router(profile.router)
 app.include_router(event.router)
 app.include_router(agenda.router)
 app.include_router(email.router)
 app.include_router(invitation.router)
 
 
+# I add the quote day to the home page
```
ivarshav, line 27: Great comment! you can add a `TODO:` prefix for it.

tests/client_fixture.py
```diff
@@ -51,6 +52,18 @@ def profile_test_client():
     Base.metadata.drop_all(bind=test_engine)
 
 
+@pytest.fixture(scope="session")
+def home_test_client():
```
ivarshav, line 56: It's a bit pity that `agenda_test_client`,  `profile_test_client` and this are very similar.
I wonder if there is a pritter way for that?

tests/test_load_quotes.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+# Tests for loading the quotes in the db:
+def test_load_daily_quotes(session):
+    load_quotes.load_daily_quotes(session)
+    assert session.query(Quote).count() > 0
+
+
+def test_load_daily_quotes_with_json_valueerror(mocker, session):
+    mock_json_load = mocker.patch('json.load')
```
ivarshav, line 12: I think you can do something like to save a line:
```
mock_json_load = mocker.patch('json.load', side_effect = ValueError)
```

tests/test_load_quotes.py
```diff
@@ -0,0 +1,22 @@
+from app.database.models import Quote
+from app.internal.quotes import load_quotes
+
+
+# Tests for loading the quotes in the db:
+def test_load_daily_quotes(session):
+    load_quotes.load_daily_quotes(session)
+    assert session.query(Quote).count() > 0
```
ivarshav, line 8: assert is_quotes_table_empty is False

tests/test_load_quotes.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_load_daily_quotes_with_json_valueerror(mocker, session):
+    mock_json_load = mocker.patch('json.load')
+    mock_json_load.side_effect = ValueError
+    load_quotes.load_daily_quotes(session)
+    assert session.query(Quote).count() == 0
```
ivarshav, line 15: assert is_quotes_table_empty() is true

app/internal/quotes/load_quotes.py
```diff
@@ -0,0 +1,41 @@
...
+    quotes_objects = [
+        Quote(text=quote['text'], author=quote['author'])
+        for quote in all_quotes
+        ]
+    session.add_all(quotes_objects)
+    session.commit()
+
+
+def is_quotes_table_empty(session: Session) -> bool:
+    return session.query(Quote).count() == 0
```
ivarshav, line 34: Maybe it's usefull to add a function that returns session.query(Quote).count()?

tests/test_quotes.py
```diff
@@ -0,0 +1,33 @@
...
+DATE2 = date(2021, 1, 2)
+
+
+def insert_quotes(session):
+    quote = Quote(
+        id=1, text='You have to believe in yourself.', author='Sun Tzu')
+    quote2 = Quote(id=2, text='Wisdom begins in wonder.', author='Socrates')
+    session.add(quote)
+    session.add(quote2)
+    session.commit()
```
ivarshav, line 16: When you delete them from db?

app/database/models.py
zohary89: No, you're right!

tests/test_quotes.py
zohary89: It's will be deleted automatically, after the session is over. (According to the fixture- 'session' in conftest).

ivarshav: I'm pretty sure you need to add something like `session.delete(quote)` in order to delete it.
Fixtures are magical, but there is a limit to magic :)

tests/test_load_quotes.py
zohary89: I thought it's better not to use an inner function of the module, for the actual testing of that module...
However, I will add separate count function for all the tests that use the count (in the tests file).

app/database/models.py
```diff
@@ -74,3 +74,11 @@ def __repr__(self):
...
             f'to {self.recipient})>'
         )
+
+
+class Quote(Base):
+    __tablename__ = "quotes"
+
+    id = Column(Integer, primary_key=True, index=True)
+    text = Column(String, nullable=False)
+    author = Column(String)
```
ivarshav, line 84: Can this property be null? :)

tests/test_load_quotes.py
```diff
@@ -0,0 +1,25 @@
+from app.database.models import Quote
+from app.internal.quotes import load_quotes
+
+
+def get_quotes_amount(session):
```
ivarshav, line 5: Maybe add that as part of the table class, and then you can use it in `is_quotes_table_empty` too.
In the tests you can do something like:
```
assert is_quotes_table_empty  # count == 0
assert not is_quotes_table_empty # count > 0
```

app/database/models.py
zohary89: Yes...

app/internal/quotes/daily_quotes.py
```diff
@@ -0,0 +1,23 @@
...
+def quote_per_day(
+        session: Session, date: date = date.today()
+        ) -> Optional[Quote]:
+    """This function provides a daily quote, relevant to the current
+    day of the year. The quote is randomally selected from a set
+    of quotes matching to the given day"""
+    day_num = date.timetuple().tm_yday
+    quotes = session.query(Quote).filter(
+        Quote.id % TOTAL_DAYS == day_num).all()
+    if len(quotes) > 0:
```
yammesicka, line 21: A more Pythonic approach: `if not quotes`

app/internal/quotes/daily_quotes.py
```diff
@@ -0,0 +1,23 @@
...
+
+def quote_per_day(
+        session: Session, date: date = date.today()
+        ) -> Optional[Quote]:
+    """This function provides a daily quote, relevant to the current
+    day of the year. The quote is randomally selected from a set
+    of quotes matching to the given day"""
+    day_num = date.timetuple().tm_yday
+    quotes = session.query(Quote).filter(
+        Quote.id % TOTAL_DAYS == day_num).all()
```
yammesicka, line 20: Maybe we should choose at random directly from the database [using the ORM](https://stackoverflow.com/questions/60805/getting-random-row-through-sqlalchemy).

app/internal/quotes/load_quotes.py
```diff
@@ -0,0 +1,41 @@
+import json
+from typing import Dict, List
+
+from app.database.models import Quote
+
+from sqlalchemy.orm import Session
+
+
```
yammesicka, line 8: Add annotations for the dictionary :)

yammesicka (changes requested):

tests/test_home.py
```diff
@@ -1,6 +1,6 @@
 class TestHome:
     URL = "/"
 
-    def test_get_page(self, client):
-        resp = client.get(self.URL)
+    def test_get_page(self, home_test_client):
+        resp = home_test_client.get(self.URL)
         assert resp.status_code == 200
```
PureDreamer, line 6: you could change it to assert resp.status_code.ok

app/templates/home.html
```diff
@@ -6,4 +6,15 @@
     <h1>{{message}}</h1>
 </div>
 
-{% endblock %}
\ No newline at end of file
+<!-- I chose to add the quote day to the home page until the relavent calendar view will be developed.  -->
```
PureDreamer, line 9: i would suggest adding a demo html for adding it because in the future the daily quote might not be there.

app/internal/quotes/load_quotes.py
```diff
@@ -0,0 +1,41 @@
...
+def get_quotes_from_json() -> List[Dict[str, Optional[str]]]:
+    """This function reads all of the daily quotes from a specific JSON file.
+    The JSON file content is copied from the free API:
+    'https://type.fit/api/quotes'. I saved the content so the API won't be
```
PureDreamer, line 12: Id suggest adding docs in the Google doc string style https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
also, there are some typos in the text.
Overall excellent job!

app/templates/home.html
zohary89: Thanks for the comment!
Since its just a simple sentence, which changes every day, I see no reason to upload a demo page. I think it will eventually appear in the main view of the calendar, but anyway I'll follow up on it and make sure to move it accordingly...

app/internal/quotes/load_quotes.py
zohary89: thanks! :)

app/main.py
```diff
@@ -14,6 +16,8 @@
 app.mount("/static", StaticFiles(directory=STATIC_PATH), name="static")
 app.mount("/media", StaticFiles(directory=MEDIA_PATH), name="media")
 
+load_quotes.load_daily_quotes(next(get_db()))
```
yammesicka, line 19: Prefer to use dependency injection using FastAPI's `Depends`

tests/client_fixture.py
```diff
@@ -51,6 +52,18 @@ def profile_test_client():
     Base.metadata.drop_all(bind=test_engine)
 
 
+@pytest.fixture(scope="session")
```
yammesicka, line 55: Maybe we should use the general test client? :)

tests/test_home.py
```diff
@@ -1,6 +1,9 @@
+from fastapi import status
+
+
 class TestHome:
     URL = "/"
 
-    def test_get_page(self, client):
-        resp = client.get(self.URL)
-        assert resp.status_code == 200
+    def test_get_page(self, home_test_client):
```
yammesicka, line 7: I prefer to leave it as it is

tests/test_quotes.py
```diff
@@ -0,0 +1,33 @@
+from datetime import date
+
+from app.database.models import Quote
+from app.internal.quotes import daily_quotes
+
+DATE = date(2021, 1, 1)
+DATE2 = date(2021, 1, 2)
+
+
+def insert_quotes(session):
```
yammesicka, line 10: Maybe it would be better to add this as a fixture? :)

yammesicka (changes requested):
Great work, very nice code! I've added few of my insights :)

app/main.py
zohary89: As I understand, it is not possible to use "depends" outside a route function. In my case, I load the data at app initialization, regardless of any request. Therefore, I need to get a db-session, which is not from "depends"..

tests/client_fixture.py
zohary89: I used get_db in 'home' function (because I added the quote to the home page), so I needed to have a specific test client for home to replace the get_db with the get_test_db. Also refers to the next comment- I could not use the regular test client and had to change the test code.

I don't know yet what will be appeared at the home page, but I wanted that when the user enters the main calendar page he will see the quote, so I chose to display it at 'home' for now.. Do you think I should move it?
Thank you!

app/main.py
yammesicka: Ah, I see. OK Then :)

tests/quotes_fixture.py
```diff
@@ -0,0 +1,37 @@
+import pytest
```
yammesicka, line 1: Great job here! :)

yammesicka (approved):

tests/client_fixture.py
yammesicka: > I used get_db in 'home' function (because I added the quote to the home page), so I needed to have a specific test client for home to replace the get_db with the get_test_db. Also refers to the next comment- I could not use the regular test client and had to change the test code.

Sababa, but I see there are already 3 similar code segments :$
Can you extract most of the code to a single function and refactor all the functions to just call it?

yammesicka: > I don't know yet what will be appeared at the home page, but I wanted that when the user enters the main calendar page he will see the quote, so I chose to display it at 'home' for now.. Do you think I should move it?

That's OK, the `home` page will be refactored soon

yammesicka (changes requested):

tests/test_home.py
```diff
@@ -1,6 +1,9 @@
+from fastapi import status
+
+
 class TestHome:
     URL = "/"
 
     def test_get_page(self, client):
         resp = client.get(self.URL)
-        assert resp.status_code == 200
+        assert resp.status_code == status.HTTP_200_OK
```
yammesicka, line 9: Prefer `.ok`

Issue #34, opened 2021-01-12
Add an option of delete event


efratush:
Allow the event creator to delete the event.
A delete button should be added for the creator only.
After pressing the delete button an "Are you sure?" dialog should be shown.
An email should be sent to all recipients of the event, informing them that the event has been deleted.

PR #129, opened 2021-01-25
Feature/delete event

efratush:
#34 
Add an event delete function + tests


efratush:
Mistake in brunch, main instead of develop.

PR #134, opened 2021-01-26
feat: International Time Coordinator **IRRELEVANT**

TamarBerger:
Description:
- Creates new 'Country' table in the database which contains 'name' column and 'timezone' column.
- Displays a Globe icon.
- Clicking on the Globe opens a filter search list that allows the user to choose the wanted country.
- The function is using the user's ip address to determine his geographical location.
- The function connects the chosen country with its matching time zone based on the db's information.
- The function receives the future meeting datetime (temporarily static) and displays the time converted to the chosen location local time.

 fix: 
1. **currently using fabricated datetime for the meeting because there is no connection to event creation.**
2. the users ip is given in the '/global_time/choose' function and required in the '/global_time/{country_name}/{chosen_datetime}/{ip}' in order to support the tests.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
yammesicka, line 1: Remove this file using `git rm`.

app/main.py
```diff
@@ -20,10 +27,114 @@
 app.include_router(email.router)
 
 
+def add_countries_to_db():
```
yammesicka, line 30: Please don't add code to the `main.py` file :)

app/main.py
```diff
@@ -20,10 +27,114 @@
...
+        request: Request,
+        session=Depends(get_db)):
+    """
+    Displays the list of all countries name from "Country" table.
+    By Click on country name the page transfers to:
+    "/global_time/{country_name}/{chosen_datetime}".
+    *** Temporarily  returns a fictitious meeting datetime. ***
+    """
+    data = session.query(Country.name).all()
+    temp_meeting_datetime = '2021-02-10 19:00:00'
```
yammesicka, line 95: Maybe we can use `datetime.datetime.now()` or something :)

app/main.py
```diff
@@ -20,10 +27,114 @@
...
+    ip = request.client.host
```
yammesicka, line 96: I'm not sure we actually want to reveal the user's IP

app/templates/event/partials/global_time.html
```diff
@@ -0,0 +1,48 @@
...
+{% block content %}
+
+<a href="/global_time/choose"><img src={{ url_for('static', path='global_timing/worldwide.png') }} alt="Globe-icon-to-choose-country" title="Choose Country"></a>
+
+{% if country is not none and time is not none %}
+    <h6>{{country}} {{time}}</h6>
+{% endif %}
+
+{% if data %}
+<input type="text" id="myInput" onkeyup="myFunction()" placeholder="Search..." title="Type in a name">
```
yammesicka, line 12: Prefer using a datalist :)

app/main.py
```diff
@@ -20,10 +27,114 @@
...
+    (The list is located in app/database/countries.py)
+    Names are described either as:
+    "Country Name, City Name" or
+    "Country Name" solely.
+    Timezones are described as "Continent/ City Name"
+    for example:
+        name: Israel, Jerusalem
+        timezone: Asia/Jerusalem
+    """
+    session = SessionLocal()
```
yammesicka, line 43: Prefer using `get_db` using `Depends`

app/main.py
```diff
@@ -20,10 +27,114 @@
...
+    for country in countries:
+        partial_name = country['name']
+        for capital in country['timezones']:
+            capital_name = capital.split('/')[-1]
+            if partial_name != capital_name:
+                name = partial_name + ', ' + capital_name
+            else:
+                name = capital_name
+            existing = session.query(Country).filter_by(name=name).first()
```
yammesicka, line 52: This might take long time to execute due to the large amount of queries we perform here.

Prefer using `on conflict ignore` to simplify this part of the code, and to get better performance.

app/main.py
```diff
@@ -20,10 +27,114 @@
...
-async def home(request: Request):
+async def home(
+        request: Request):
     return templates.TemplateResponse("home.html", {
         "request": request,
         "message": "Hello, World!"
+    })
+
 
+@app.get("/global_time")
```
yammesicka, line 72: This code should be on /routers/times.py

app/main.py
```diff
@@ -20,10 +27,114 @@
...
+    """
+    Displays a globe icon.
+    By Click the page transfers to "/global_time/choose".
+    """
+    return templates.TemplateResponse("event/partials/global_time.html", {
+        "request": request
     })
+
+
+@app.get("/global_time/choose")
```
yammesicka, line 84: Maybe /time/countries? :)

app/main.py
```diff
@@ -20,10 +27,114 @@
...
+async def time_choose(
+        request: Request,
+        session=Depends(get_db)):
+    """
+    Displays the list of all countries name from "Country" table.
+    By Click on country name the page transfers to:
+    "/global_time/{country_name}/{chosen_datetime}".
+    *** Temporarily  returns a fictitious meeting datetime. ***
+    """
+    data = session.query(Country.name).all()
```
yammesicka, line 94: I think this one should be on a separate function, cached by `@functools.lru_cache`.

The countries probably won't change, and querying the database every time can to be a bit expensive :)

app/main.py
yammesicka: Also, prefer defining an `APIRoute` with `/time` as the prefix :timer_clock:

app/templates/event/partials/global_time.html
```diff
@@ -0,0 +1,48 @@
...
+        {% for name in data %}
+            <li><a href='/global_time/{{name[0]}}/{{date}}/{{ip}}'>{{ name[0] }}</a></li>
+        {% endfor %}
+    </ul>
+</div>
+
+
+<script>
+function myFunction() {
+    var input, filter, ul, li, a, i, txtValue;
```
yammesicka, line 25: Prefer using `let`

app/templates/event/partials/global_time.html
```diff
@@ -0,0 +1,48 @@
...
+    <ul id="myUL">
+        {% for name in data %}
+            <li><a href='/global_time/{{name[0]}}/{{date}}/{{ip}}'>{{ name[0] }}</a></li>
+        {% endfor %}
+    </ul>
+</div>
+
+
+<script>
+function myFunction() {
```
yammesicka, line 24: Please use better names :)

app/templates/event/partials/global_time.html
```diff
@@ -0,0 +1,48 @@
...
+
+{% if country is not none and time is not none %}
+    <h6>{{country}} {{time}}</h6>
+{% endif %}
+
+{% if data %}
+<input type="text" id="myInput" onkeyup="myFunction()" placeholder="Search..." title="Type in a name">
+
+<div class="countries_list_container"> 
+    <ul id="myUL">
```
yammesicka, line 15: Please name the id and the classes better

app/templates/event/partials/global_time.html
yammesicka, line 12: Prefer using `addEventListener` and not attaching `onkeyup` to the HTML element

yammesicka (changes requested):

yammesicka:
Great work! :) I've added few of my insights, please take a look.

app/main.py
TamarBerger: I started by using datetime.datetime.now() but I wanted to give an example for future date. 
It is just a temporary substitute until we will have real meetings data to use:)

PR #142, opened 2021-01-28
Feature/astronomy

hadaskedar2020:
This feature gets astronomical data for given date & location.
It returns the following:
Status - success / failure (+ ErrorDescription).
Location - relevant location values: name, region, country, lat, lon etc.
Astronomical data in local time: sunrise, sunset, moonrise, moonset, moon_phase, moon_illumination.


app/internal/astronomy.py
```diff
@@ -0,0 +1,88 @@
...
+        error_text (str) - error message.
+    """
+    input_query_string = dict(key=config.ASTRONOMY_API_KEY, q=location,
+                              dt=formatted_date)
+    try:
+        response = requests.request("GET", ASTRONOMY_URL,
+                                    params=input_query_string)
+    except requests.exceptions.RequestException:
+        return None, NO_API_RESPONSE
+    if response.ok:
```
yammesicka, line 54: Prefer to flip this and [return early](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement)

app/internal/astronomy.py
```diff
@@ -0,0 +1,88 @@
...
+    """ get the relevant astronomical data by calling the "weather api" API.
+    Args:
+        formatted_date (date) - relevant date.
+        location (str) - location name.
+    Returns:
+        response_json (json dict) - relevant part (data / error) of the
+        JSON returned by the API.
+        error_text (str) - error message.
+    """
+    input_query_string = dict(key=config.ASTRONOMY_API_KEY, q=location,
```
yammesicka, line 47: Prefer dictionary literal: `{'key': config.ASTRONOMY_API_KEY, 'q': ....}`

app/internal/astronomy.py
```diff
@@ -0,0 +1,88 @@
...
+                              dt=formatted_date)
+    try:
+        response = requests.request("GET", ASTRONOMY_URL,
```
yammesicka, line 50: Prefer to work with async library (like httpx)

app/internal/astronomy.py
```diff
@@ -0,0 +1,88 @@
...
+                      if isinstance(arg, dict) else arg for arg in args])
+        kwargs = {k: frozendict.frozendict(v) if isinstance(v, dict) else v
+                  for k, v in kwargs.items()}
+        return func(*args, **kwargs)
+    return wrapped
+
+
+@freezeargs
+@functools.lru_cache(maxsize=128, typed=False)
+def get_data_from_api(formatted_date: str, location: str)\
```
yammesicka, line 36: I think we should prefer returning a dictionary: 
`{'success': True, 'astronomy': ...}`

Or:
`{'success': False, 'reason': requests.reason}`

You could also `raise` an exception and catch it in the caller function.

app/config.py.example
```diff
@@ -27,6 +27,7 @@ PICTURE_EXTENSION = '.png'
 AVATAR_SIZE = (120, 120)
 
 # API-KEYS
+ASTRONOMY_API_KEY = os.getenv('ASTRONOMY_API_KEY')
```
yammesicka, line 30: A short explanation here about how to create the API Key might be helpful :)

app/internal/astronomy.py
```diff
@@ -0,0 +1,88 @@
...
+    def wrapped(*args, **kwargs):
+        args = tuple([frozendict.frozendict(arg)
+                      if isinstance(arg, dict) else arg for arg in args])
+        kwargs = {k: frozendict.frozendict(v) if isinstance(v, dict) else v
+                  for k, v in kwargs.items()}
+        return func(*args, **kwargs)
+    return wrapped
+
+
+@freezeargs
```
yammesicka, line 34: Why we need the freezeargs?

yammesicka (changes requested):

app/internal/astronomy.py
hadaskedar2020: After following your advices it seems that it is no longer necessary.
I've removed it :-)

app/internal/astronomy.py
hadaskedar2020: I've changed to asyncio and the code runs fine from the main program,
But the test checks (which runs perfectly on my computer) just fails here:
[https://github.com/PythonFreeCourse/calendar/pull/142/checks?check_run_id=1799100569](url)
Couldn't find and information regarding to what causes the problem...
Any advice?

yammesicka: I think that's because you mock using `responses` which is more suitable to `requests` than to `httpx`

hadaskedar2020: Thanks. You're right of course. Now I've changed it and it seems that my tests are fine and yet, there is an error in: tests/test_whatsapp.py::test_end_to_end_testing ERROR
[https://github.com/PythonFreeCourse/calendar/pull/142/checks?check_run_id=1809067309](url)
I've no idea on how my code could impact that...
Can you take a look again please?

yammesicka: I think it some caveats in the work with async frameworks.

Can you please `return get_data_from_api(formatted_date, location)` instead of putting it in a async loop?

hadaskedar2020: Since I use async and httpx I think have to perform the loop - is there another way?
I've tried to do it prior to the return command, but still the whatsapp test fails...

yammesicka (approved):

PR #156, opened 2021-01-30
Feature/salary calculator

ShiZinDle:
New feature - Salary Calculator:
allows users to select any category they previously created and create salary calculation settings for it. the feature also adds edit pages for the settings and view pages for the calculations. The calculation will be done for the chosen month and will be based on all events that are labeled with the corresponding category. Multiple salary settings are allowed, but only 1 for each user and category combination.

Further changes implemented:
-moved contents of file "app.database.database" to "app.database" for easier access.
-moved function "app.database,database.get_db" to "app.dependencies.get_db".
-merged file "tests.utils" into "app.internal.utils" to prevent code duplication.
-fixed imports and type hints where necessary.


PureDreamer:
In order to pass the tests you need the files to be ok with flake8, you can choose flake8 as a linter, and it will show you where are your mistakes.

ShiZinDle:
All linting errors, code conflicts and failing tests have been fixed and resolved.

app/routers/salary/utils.py
```diff
@@ -0,0 +1,494 @@
...
+        date = end.date()
+    try:
+        return (datetime.combine(date, time(0)),
+                datetime.combine(date + timedelta(1),
+                                 time(0)))
+    except NameError:
+        return datetime.min, datetime.min
+
+
+def get_total_synchronous_hours(
```
yammesicka, line 132: How about [this answer](https://stackoverflow.com/a/9044111/1058671)? :)

app/routers/salary/utils.py
```diff
@@ -0,0 +1,494 @@
...
+        float: The weekly overtime amount for the supplied shifts.
+
+    Raises:
+        None
+    """
+    total_week_hours = sum(get_shift_len(shift.start, shift.end)
+                           for shift in shifts)
+    if total_week_hours <= wage.week_working_hours:
+        return 0.0
+    else:
```
yammesicka, line 278: The `else:` is redundant here

app/routers/salary/utils.py
```diff
@@ -0,0 +1,494 @@
...
+    if is_night_shift(start, end, wage):
+        hour_basis = wage.night_hour_basis
+    else:
+        hour_basis = wage.regular_hour_basis
+    shift_len = get_shift_len(start, end)
+
+    total_hours += get_hours_during_holiday(start, end, wage) / 2
+
+    if shift_len > hour_basis:
+        overtime += shift_len - hour_basis
```
yammesicka, line 222: This should be a separate function :)

app/routers/salary/utils.py
```diff
@@ -0,0 +1,494 @@
...
+
+    Args:
+        user_id (int): Id of the relevant user.
+        category_id (int): Id for the relevant category.
+
+    Returns:
+        SalarySettings | None: Settings for the provided user_id and
+                               category_id if exists, None otherwise.
+    """
+    session = SessionLocal()
```
yammesicka, line 470: Prefer working with get_db dependency

yammesicka (changes requested):
Wow! That's some effort you put in this PR. Awesome job!

app/routers/salary/config.py
```diff
@@ -0,0 +1,25 @@
+from datetime import time
```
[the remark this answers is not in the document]
ShiZinDle: Not sure how that happened, will resolve.

app/database/models.py
```diff
@@ -110,6 +125,78 @@ def __repr__(self):
...
+class SalarySettings(Base):
+    # Code revision required after categories feature is added
+    # Code revision required after holiday times feature is added
+    # Code revision required after Shabbat times feature is added
+    __tablename__ = "salary_settings"
+
+    user_id = Column(
+        Integer, ForeignKey("users.id"), primary_key=True,
+    )
+    # category_id = Column(
```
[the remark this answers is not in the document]
yammesicka: Cool, marking as solved

app/internal/search.py
```diff
@@ -16,7 +16,7 @@ def get_stripped_keywords(keywords: str) -> str:
 
 
 def get_results_by_keywords(
-                            session: SessionLocal,
+                            session: Session,
```
yammesicka, line 19: Please use Depends(get_db) (also in other files, if applicable)

tests/salary/conftest.py
```diff
@@ -0,0 +1,79 @@
...
+    'create_settings': 'Already created your settings?',
+    'pick_settings': 'Edit Settings',
+    'edit_settings': 'Settings don\'t need editing?',
+    'pick_category': 'View Salary',
+    'view_salary': 'Need to alter your settings?',
+    'salary_calc': 'Total Salary:',
+}
+
+ROUTES = {
+    'home': '/salary',
```
yammesicka, line 23: Use url_for instead :)

yammesicka (changes requested):
Awesome job, added few additional insights

app/internal/search.py
ShiZinDle: I don't think i fully understand your request here.
As I've mentioned before, you can see in [This comment by Tiangolo](https://github.com/tiangolo/fastapi/issues/1105#issuecomment-609919850), that Depends is not meant for use outside of path operation. Additionally, this function doesn't call a session directly, but is rather passed one from a path operation (where it is created using Depends(get_db)).
I should also mentioned that this is not part of my code and I merely fixed the incorrect type hinting in this module.

yammesicka: I see, you are right. Sababa, let's keep it this way :)

yammesicka (changes requested):
Awesome, a little fix and it's ready to merge from my side. Please also solve the conflicts :)

app/routers/salary/utils.py
```diff
@@ -0,0 +1,544 @@
...
+        category_id (int): Id for the relevant category.
+
+    Returns:
+        SalarySettings | None: Settings for the provided user_id and
+                               category_id if exists, None otherwise.
+    """
+    settings = session.query(SalarySettings).filter(and_(
+        SalarySettings.user_id == user_id, SalarySettings.category_id
+        == category_id)).first()
+    session.close()
```
yammesicka, line 488: Is this necessary?

ShiZinDle: I can't really say why, but for some reason removing this line fails one of my tests.
The functionality is essentially the same, and the service itself doesn't seem to be affected, yet I didn't manage to pass the tests without it.
If you want to take a look at it, the one that fails is "test_routes: :test_edit_settings".

yammesicka: Try to `git checkout upstream/develop -- requirements.txt` and `pip install -U -r requirements.txt`.
We've downgraded the pytest-asyncio library, that's might be the problem

ShiZinDle: Unfortunately this did not solve the problem.

yammesicka (approved):

yammesicka:
While trying to solve conflicts, I've seen there is a minor problem with test_events.py. Can you please take a look? :)

ShiZinDle:
> While trying to solve conflicts, I've seen there is a minor problem with test_events.py. Can you please take a look? :)

There was a missing import, now fixed.

PR #158, opened 2021-01-30
Bugfix/logging system

Liad-n:
Disconnected the logger from the app instance due to problematic circular imports when used from within routes. 
Now to use the logger, 

```from loguru import logger```

and use it this way:

logger.debug(message)
logger.error(message)
...


yammesicka (approved):
Great job!

PR #159, opened 2021-01-30
Feat: Event View Backend

OdeChan:
Improved backend of the get method endpoint of "eventview" to show actual event details instead of placeholders.

On the way:
- Fixed up get_event_by_id (previously by_id), including error handling
- Fixed delete_event and update_event to fit new get_event_by_id function
- Moved logger to dependencies so can be used outside of main
- Fixed up the event_test_client
- Added and fixed up event tests


app/routers/event.py
```diff
@@ -25,16 +27,64 @@ async def eventedit(request: Request):
...
+@router.get("/view/{event_id}")
+async def eventview(request: Request, event_id: int,
+                    db: Session = Depends(get_db)):
+    try:
+        event = get_event_by_id(db, event_id)
+    except NoResultFound:
+        raise HTTPException(status_code=404, detail="Event not found")
+    except MultipleResultsFound:
+        raise HTTPException(status_code=500, detail="Multiple events found")
+    start_format = '%A, %d/%m/%Y %H:%M'
```
sagizaidor, line 39: למה יש גם את שם החודש וגם את מספר החודש?

app/routers/event.py
```diff
@@ -113,38 +166,10 @@ def get_participants_emails_by_event(db: Session, event_id: int) -> List[str]:
     """Returns a list of all the email address of the event invited users,
         by event id."""
 
-    return [email[0] for email in db.query(User.email).
+    return (
+        [email[0] for email in db.query(User.email).
             select_from(Event).
```
sagizaidor, line 171: תהיה: אולי שווה לפתוח על זה טבלה בDB?

sagizaidor (commented):
נראה אחלה,
לא היו יותר מידי דברים שתפסו את תשומת ליבי

app/routers/event.py
OdeChan: @sagizaidor %A is the is the week day, %m is the month and %M is minutes. Am I missing something?

app/routers/event.py
OdeChan: @sagizaidor I'm not actually sure, this isn't my code, but lint check failed me for it so I had to fix

app/routers/event.py
sagizaidor: אוי...תכלס, טעות שלי, התבלבלתי עם %B,...

app/routers/event.py
```diff
@@ -25,16 +27,66 @@ async def eventedit(request: Request):
...
+        event = get_event_by_id(db, event_id)
+    except NoResultFound:
+        raise HTTPException(status_code=404, detail="Event not found")
+    except MultipleResultsFound:
+        raise HTTPException(status_code=500, detail="Multiple events found")
+
+    participants = get_participants_emails_by_event(db, event_id)
+
+    db.delete(event)  # Delete event
+    db.query(UserEvent).filter_by(event_id=event_id) \
```
yammesicka, line 62: Prefer not to use backslash for the split. In this case, you can merge the lines. (You can remove the comment)

app/routers/event.py
```diff
@@ -25,16 +27,66 @@ async def eventedit(request: Request):
...
+    if not isinstance(db, Session):
+        raise AttributeError(
+            f'Could not connect to database. '
+            f'db instance type received: {type(db)}')
+    try:
+        event = db.query(Event).filter_by(id=event_id).one()
+    except NoResultFound:
+        raise NoResultFound(f"Event ID does not exist. ID: {event_id}")
+    except MultipleResultsFound:
+        error_message = f'Multiple results found when getting event. ' \
```
yammesicka, line 85: Prefer parenthesis here

app/routers/event.py
```diff
@@ -25,16 +27,66 @@ async def eventedit(request: Request):
                                       {"request": request})
 
 
-@router.get("/view/{id}")
-async def eventview(request: Request, id: int):
+@router.get("/view/{event_id}")
```
yammesicka, line 30: Is `/view/` needed?

yammesicka (changes requested):

app/templates/event/partials/view_event_details_tab.html
```diff
@@ -1,32 +1,30 @@
...
 </div>
 
-<div class="event_info_row">
-    <span class="icon">ICON</span>
-    <span>Repeats every INTERVAL</span>
-</div>
+<!--<div class="event_info_row">-->
+<!--    <span class="icon">ICON</span>-->
+<!--    <span>Repeats every INTERVAL</span>-->
+<!--</div>-->
```
ron-huberfeld, line 21: האם זה תורם להשאיר את ההערות בקוד?

OdeChan: Yes, these are placeholders till there are matching model fields

app/routers/event.py
efratush: לכל משתמש יש אימייל ויש טבלה מקשרת בין האירוע למשתמשים. איזה טבלה אתה מציע להוסיף?

app/routers/event.py
OdeChan: Probably not particularly. I just like the uniformity, but I guess it could be removed

OdeChan: @yammesicka Probably not particularly. I just like the uniformity, but I guess it could be removed

app/routers/event.py
```diff
@@ -25,16 +27,66 @@ async def eventedit(request: Request):
...
+async def eventview(request: Request, event_id: int,
+                    db: Session = Depends(get_db)):
+    try:
+        event = get_event_by_id(db, event_id)
+    except NoResultFound:
+        raise HTTPException(status_code=404, detail="Event not found")
+    except MultipleResultsFound:
```
efratush, line 37: There can be no multiplication id in db

OdeChan: True, I used the exception in case something went terribly wrong and somehow an event id did duplicate. When using .one() this will throw this exception. Using .first() would have return an arbitrary one, which would be more dangerous as it would potentially show the wrong event to the user

app/routers/event.py
```diff
@@ -25,16 +27,66 @@ async def eventedit(request: Request):
...
+        raise HTTPException(status_code=500, detail="Multiple events found")
+
+    participants = get_participants_emails_by_event(db, event_id)
+
+    db.delete(event)  # Delete event
```
efratush, line 61: Why did you remove Try?

app/routers/event.py
```diff
@@ -25,16 +27,66 @@ async def eventedit(request: Request):
...
+    """Gets a single event by id"""
+    if not isinstance(db, Session):
+        raise AttributeError(
+            f'Could not connect to database. '
+            f'db instance type received: {type(db)}')
+    try:
+        event = db.query(Event).filter_by(id=event_id).one()
+    except NoResultFound:
+        raise NoResultFound(f"Event ID does not exist. ID: {event_id}")
+    except MultipleResultsFound:
```
efratush, line 84: There can be no multiplication id in db

app/routers/event.py
efratush: This is not a possible situation on the part of db

app/routers/event.py
OdeChan: I moved it onto get_event_by_id, it will make sure that the event exists before trying to delete it. If I missed an exception, let me know :)

app/routers/event.py
OdeChan: Answered above

app/routers/event.py
efratush: Maybe because there is nothing general that will catch, it is right to leave the catch

app/routers/event.py
OdeChan: Ok, I think I fixed it no, sorry :)

yammesicka (approved):

PR #165, opened 2021-01-30
feat: adding holidays to calendar as events

galzunited:
https://forums.pythonic.guru/t/topic/8637

Added an option for user to import holidays from ics file.
Took the naïve approach for now - not checking for duplicates while importing.
Need to think about this mechanism - since checking every date to see if holiday is there is too expensive.
As for the first step i think its ok (:


tests/test_holidays.py
```diff
@@ -0,0 +1,34 @@
...
+    def create_user(session):
+        user = User(username='testuser', email='test@abc.com', password='1234')
+        session.add(user)
+        session.commit()
+        return user
+
+    @staticmethod
+    def test_import_holidays_page_exists(client):
+        resp = client.get(TestHolidaysImport.HOLIDAYS)
+        assert resp.status_code == status.HTTP_200_OK
```
yammesicka, line 22: Prefer `.ok`

tests/test_holidays.py
```diff
@@ -0,0 +1,34 @@
...
+from app.routers import profile
+from fastapi import status
+import os
+
+
+class TestHolidaysImport:
+    HOLIDAYS = '/profile/import_holidays'
+
+    @staticmethod
+    def create_user(session):
```
yammesicka, line 13: Please use a fixture that also deletes the user on teardown

app/resources/ics_example.txt
```diff
@@ -0,0 +1,36 @@
+STATUS:CONFIRMED
```
yammesicka, line 1: Prefer to put this in internal folder inside  `tests`

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+            userevent = UserEvent(
+                user_id=holiday.owner_id,
+                event_id=holiday.id
+            )
+            userevents.append(userevent)
+        session.add_all(userevents)
+        session.commit()
+
+    except Exception as ex:
+        print(ex)
```
yammesicka, line 223: Use logging instead

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+    return holidays
+
+
+async def save_holidays_to_db(holidays, session):
+    """
+    this function saves holiday list into database.
+    :param holidays: list of holidays events
+    :param session: current connection
+    """
+    try:
```
yammesicka, line 208: Catch only the things that might fail

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+        for holiday in holidays:
+            userevent = UserEvent(
+                user_id=holiday.owner_id,
+                event_id=holiday.id
+            )
+            userevents.append(userevent)
+        session.add_all(userevents)
+        session.commit()
+
+    except Exception as ex:
```
yammesicka, line 222: Try not to catch broad `Exception` type - catch more specific exception types

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+    :return:list of holidays events
+    """
+    regex = re.compile(
+        r'SUMMARY:(?P<title>.*)(\n.*){1,8}DTSTAMP:(?P<date>\w{8})',
+        re.MULTILINE)
+    holidays_iter = regex.finditer(file)
+    holidays = []
+    for holiday in holidays_iter:
+        title = holiday.groupdict()['title'].strip()
+        title_to_save = ''.join([i if ord(i) < 128 else '' for i in title])
```
yammesicka, line 189: `title_to_save = ''.join(i for i in title if ord(i) < 128)`

Also, put 128 in a variable (The reader might ask himself: Why 128? What does it means?)

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+    and date from standrd ics file
+    :param file:standard ics file
+    :param session:current connection
+    :return:list of holidays events
+    """
+    regex = re.compile(
+        r'SUMMARY:(?P<title>.*)(\n.*){1,8}DTSTAMP:(?P<date>\w{8})',
+        re.MULTILINE)
+    holidays_iter = regex.finditer(file)
+    holidays = []
```
yammesicka, line 186: Maybe `parsed_holidays`?

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+
+def get_holidays_from_file(file, session):
+    """
+    this function using regex to extract holiday title
+    and date from standrd ics file
+    :param file:standard ics file
+    :param session:current connection
+    :return:list of holidays events
+    """
+    regex = re.compile(
```
yammesicka, line 182: Prefer compiling it outside the function to prevent compilation every time the function is called

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+
+@router.post("/update_holidays")
+async def update_holidays(
+        file: UploadFile = File(...), session=Depends(get_db)):
+    icsfile = await file.read()
+    holidays = get_holidays_from_file(icsfile.decode(), session)
+    try:
+        await save_holidays_to_db(holidays, session)
+    except Exception as ex:
+        print(ex)
```
yammesicka, line 168: Please use our logging system

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+
+
+@router.post("/update_holidays")
+async def update_holidays(
+        file: UploadFile = File(...), session=Depends(get_db)):
+    icsfile = await file.read()
+    holidays = get_holidays_from_file(icsfile.decode(), session)
+    try:
+        await save_holidays_to_db(holidays, session)
+    except Exception as ex:
```
yammesicka, line 167: Prefer catching less broad exception type

app/routers/profile.py
```diff
@@ -148,3 +144,80 @@ def get_image_crop_area(width, height):
...
+    })
+
+
+@router.post("/update_holidays")
+async def update_holidays(
+        file: UploadFile = File(...), session=Depends(get_db)):
+    icsfile = await file.read()
+    holidays = get_holidays_from_file(icsfile.decode(), session)
+    try:
+        await save_holidays_to_db(holidays, session)
```
yammesicka, line 166: I don't think you should put `await` here :)

yammesicka (commented):
Great job! I've added few of my insights :)

app/routers/profile.py
galzunited: i read in [SO](https://stackoverflow.com/questions/3427329/compiling-a-regex-inside-a-function-thats-called-multiple-times) that re.compile use cache so im not sure about the difference - anyway, ill move it to global

app/routers/profile.py
```diff
@@ -148,3 +149,64 @@ def get_image_crop_area(width, height):
         return (delta, 0, width - delta, height)
     delta = (height - width) // 2
     return (0, delta, width, width + delta)
+
+
+@router.post("/update_holidays")
+async def update_holidays(
+        file: UploadFile = File(...), session=Depends(get_db)):
```
ivarshav, line 156: Is it possible to validate the file format?

app/routers/profile.py
```diff
@@ -148,3 +149,64 @@ def get_image_crop_area(width, height):
...
+    except SQLAlchemyError as ex:
+        logger.error(ex)
+    finally:
+        url = router.url_path_for("profile")
+        return RedirectResponse(url=url, status_code=HTTP_302_FOUND)
+
+
+def get_holidays_from_file(file, session):
+    """
+    this function using regex to extract holiday title
```
ivarshav, line 170: Capital T :)

app/routers/profile.py
```diff
@@ -148,3 +149,64 @@ def get_image_crop_area(width, height):
...
+    holidays = []
+    for holiday in parsed_holidays:
+        valid_ascii_chars_range = 128
+        title = holiday.groupdict()['title'].strip()
+        title_to_save = ''.join([i if ord(i) < valid_ascii_chars_range
+                                 else '' for i in title])
+        date = holiday.groupdict()['date'].strip()
+        holiday = Event(
+            title=title_to_save,
+            start=datetime.strptime(date, '%Y%m%d'),
```
ivarshav, line 186: Maybe save the format as a constat?

tests/test_holidays.py
```diff
@@ -0,0 +1,23 @@
+from app.database.models import Event
```
ivarshav, line 1: An idea for another test-case: get file with wrong format.

app/routers/profile.py
```diff
@@ -148,3 +149,64 @@ def get_image_crop_area(width, height):
...
+
+@router.post("/update_holidays")
+async def update_holidays(
+        file: UploadFile = File(...), session=Depends(get_db)):
+    icsfile = await file.read()
+    holidays = get_holidays_from_file(icsfile.decode(), session)
+    try:
+        save_holidays_to_db(holidays, session)
+    except SQLAlchemyError as ex:
+        logger.error(ex)
```
yammesicka, line 162: Prefer `logger.exception`

app/routers/profile.py
```diff
@@ -148,3 +149,64 @@ def get_image_crop_area(width, height):
...
+    this function using regex to extract holiday title
+    and date from standrd ics file
+    :param file:standard ics file
+    :param session:current connection
+    :return:list of holidays events
+    """
+    parsed_holidays = REGEX_EXTRACT_HOLIDAYS.finditer(file)
+    holidays = []
+    for holiday in parsed_holidays:
+        valid_ascii_chars_range = 128
```
yammesicka, line 179: These four lines should be in a separate function :)

yammesicka (changes requested):
Great job!

app/routers/profile.py
galzunited: saved it as a variable as its the only function which use it (for now). thanks.

app/routers/profile.py
galzunited: Do you mean to check if the file has ics extension or validate the file itself for valid ics file? for the first one we have the input accept=".ics" but i guess i could check it after upload as well. for the later, as its text file and im using regex, if i wont have the valid format im looking for ill get 0 results.

tests/test_holidays.py
galzunited: Since im using regex - wrong format will just yield zero results. ill add another test with wrong structre.

yammesicka (approved):
Suggested only a minor fix. Please add type annotations to your functions and we're good to go :)

app/routers/profile.py
```diff
@@ -158,3 +167,71 @@ def get_image_crop_area(width, height):
...
+    try:
+        save_holidays_to_db(holidays, session)
+    except SQLAlchemyError as ex:
+        logger.exception(ex)
+    finally:
+        url = router.url_path_for("profile")
+        return RedirectResponse(url=url, status_code=HTTP_302_FOUND)
+
+
+def get_holidays_from_file(file, session) -> List[Event]:
```
yammesicka, line 186: Add type annotations

app/routers/profile.py
```diff
@@ -158,3 +167,71 @@ def get_image_crop_area(width, height):
...
+    parsed_holidays = REGEX_EXTRACT_HOLIDAYS.finditer(file)
+    holidays = []
+    for holiday in parsed_holidays:
+        holiday = create_holiday_event(
+            holiday, session.query(User).filter_by(id=1).first().id)
+        holidays.append(holiday)
+    return holidays
+
+
+def create_holiday_event(holiday, owner_id) -> Event:
```
yammesicka, line 203: Add type annotations

app/routers/profile.py
```diff
@@ -158,3 +167,71 @@ def get_image_crop_area(width, height):
...
+        title=title_to_save,
+        start=datetime.strptime(date, format_string),
+        end=datetime.strptime(date, format_string) + timedelta(days=1),
+        content='holiday',
+        owner_id=owner_id
+    )
+    return holiday
+
+
+def save_holidays_to_db(holidays, session):
```
yammesicka, line 220: Add type annotations

tests/test_holidays.py
```diff
@@ -0,0 +1,35 @@
...
+from app.routers import profile
+import os
+
+
+class TestHolidaysImport:
+    HOLIDAYS = '/profile/import_holidays'
+
+    @staticmethod
+    def test_import_holidays_page_exists(client):
```
yammesicka, line 10: Add type annotations to the functions in this file :)

yammesicka (changes requested):

app/routers/profile.py
```diff
@@ -137,6 +145,13 @@ async def update_telegram_id(
     return RedirectResponse(url=url, status_code=HTTP_302_FOUND)
 
 
+@router.get("/import_holidays")
```
yammesicka, line 148: Prefer: `/holidays/import`

app/routers/profile.py
```diff
@@ -154,3 +169,71 @@ def get_image_crop_area(width, height):
         return delta, 0, width - delta, height
     delta = (height - width) // 2
     return 0, delta, width, width + delta
+
+
+@router.post("/update_holidays")
```
yammesicka, line 174: `/holidays/update`

app/templates/import_holidays.html
```diff
@@ -0,0 +1,18 @@
...
+
+
+{% block content %}
+
+<form class="m-3" action="update_holidays" method="post" enctype="multipart/form-data">
+    <h5 class="title">Import holidays using ics file</h5>
+    <h6 class="title">Use the upload button to choose ics file</h6>
+    <a href="https://www.feiertagskalender.ch/export.php?geo=3542&hl=en#" target="_blank">Check this website to export holidays by country</a>
+    <div class="input-group">
+        <input type="file" class="form-control-sm" accept=".ics" name="file">
```
yammesicka, line 11: Maybe it would be better to add a label here :)

tests/test_holidays.py
```diff
@@ -0,0 +1,36 @@
...
+        resp = client.get(TestHolidaysImport.HOLIDAYS)
+        print(resp.ok)
```
yammesicka, line 13: Remove the prints

tests/test_holidays.py
```diff
@@ -0,0 +1,36 @@
+from app.database.models import Event, User
```
yammesicka, line 1: Resort the imports

yammesicka (changes requested):
Awesome job! I've added few additional comments :)

app/templates/import_holidays.html
galzunited: not sure i understand - what kind of label?

yammesicka: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

galzunited: I'm just not sure what is the label use case for this page.

yammesicka: It improves a11y :)

app/templates/import_holidays.html
```diff
@@ -0,0 +1,18 @@
+{% extends "base.html" %}
+
+
+{% block content %}
+
+<form class="m-3" action="update_holidays" method="post" enctype="multipart/form-data">
+    <h5 class="title">Import holidays using ics file</h5>
+    <h6 class="title">Use the upload button to choose ics file</h6>
```
yammesicka, line 8: Why `h6` btw? Is that a 6th level title?

yammesicka (changes requested):

app/templates/import_holidays.html
galzunited: Thanks. good to know!

app/templates/import_holidays.html
galzunited: Used it just for font size of the heading.

yammesicka: Use CSS for this. h5/h6 has semantic meaning of a title :)

app/internal/import_holidays.py
```diff
@@ -0,0 +1,64 @@
...
+    :param file:standard ics file
+    :param session:current connection
+    :return:list of holidays events
+    """
+    parsed_holidays = REGEX_EXTRACT_HOLIDAYS.finditer(file)
+    holidays = []
+    for holiday in parsed_holidays:
+        holiday_event = create_holiday_event(
+            holiday, session.query(User).filter_by(id=1).first().id)
+        holidays.append(holiday_event)
```
yammesicka, line 26: Maybe it's better to return them as an iterator? (`yield`)

app/routers/profile.py
```diff
@@ -1,14 +1,18 @@
...
 from PIL import Image
 from starlette.responses import RedirectResponse
 from starlette.status import HTTP_302_FOUND
+from sqlalchemy.exc import SQLAlchemyError
 
 from app import config
 from app.database.models import User
 from app.dependencies import get_db, MEDIA_PATH, templates
 from app.internal.on_this_day_events import get_on_this_day_events
+from app.internal.import_holidays import get_holidays_from_file, \
```
yammesicka, line 14: Prefer splitting lines using parenthesis

app/templates/import_holidays.html
```diff
@@ -0,0 +1,20 @@
+{% extends "base.html" %}
+
+
+{% block content %}
+
+<form class="m-3" action="update_holidays" method="post" enctype="multipart/form-data">
+    <h1 class="title">Import holidays using ics file</h1>
```
yammesicka, line 7: h1?

app/templates/import_holidays.html
```diff
@@ -0,0 +1,20 @@
...
+    <p class="subtitle" style="font-size: 20px;">Use the upload button to choose ics file</p>
```
yammesicka, line 8: Prefer using CSS, avoid using `style` attribute

app/templates/import_holidays.html
```diff
@@ -0,0 +1,20 @@
...
+    <a href="https://www.feiertagskalender.ch/export.php?geo=3542&hl=en#" target="_blank">Check this website to export holidays by country</a>
+    <div class="input-group">
+        <label for="file">
```
yammesicka, line 11: Please read about how to use the `label` element

yammesicka (changes requested):

app/templates/import_holidays.html
galzunited: why not? its the title

app/internal/import_holidays.py
galzunited: Im saving the list all together, so iterator is not the right case here IMO.

yammesicka (approved):

galzunited:
Hi Yam - found out im having weird issues at the moment - when trying to add the file and import im getting 
"Did not find CR at end of boundary (40)"
While trying to import using postman everything works fine.
Tried to find for answers, might relate to multipart, but in django it seems we are using to correct enctype for the form.
Can you please take a look at it?

galzunited:
almost sure its related to the route we add - as in the first place when everything was under profile route it works fine.

PR #168, opened 2021-01-31
Telegram v2

leddest:
- Updated getting current events
- Added creating new events


app/telegram/models.py
```diff
@@ -27,6 +27,8 @@ def _get_first_name(self, data: Dict) -> str:
 
 
 class Bot:
+    MEMORY = {}
```
yammesicka, line 30: This is problematic. It's equivalent to a global variable and might create problems when working asynchronously.

app/telegram/handlers.py
```diff
@@ -76,26 +97,145 @@ async def this_week_handler(self):
...
+            text += f'Starts on:\n{start}\n\n'
+            text += f'Ends on:\n{end}'
+            await telegram_bot.send_message(
+                chat_id=self.chat.user_id, text=text)
+            await asyncio.sleep(1)
 
-        await telegram_bot.send_message(chat_id=self.chat.user_id, text=answer)
+        return answer
+
+    async def process_new_event(self, memo_dict):
```
yammesicka, line 131: Split to functions

app/telegram/handlers.py
```diff
@@ -76,26 +97,145 @@ async def this_week_handler(self):
...
+
+        answer = f"{chosen_date.strftime('%A, %B %d')}:\n"
+        await telegram_bot.send_message(
+            chat_id=self.chat.user_id, text=answer)
 
         for event in events:
-            answer += f'''
-From {event.start.strftime('%d/%m %H:%M')} \
-to {event.end.strftime('%d/%m %H:%M')}: {event.title}.\n'''
+            start = event.start.strftime("%d %b %Y %H:%M")
```
yammesicka, line 118: This should be a separate function :)

yammesicka (changes requested):
Awesome job! Looks really cool. I left you with 1 major architecture change to do, and 2 SRP issues :)

app/telegram/bot.py
```diff
@@ -9,3 +11,7 @@
 WEBHOOK_URL = settings.webhook_url
 
 telegram_bot = Bot(BOT_API, WEBHOOK_URL)
+
+loop = asyncio.get_event_loop()
```
yammesicka, line 15: I think you can just `asyncio.run` the `set_webhook`

yammesicka (changes requested):

app/telegram/bot.py
leddest: It's impossible - asyncio.run() cannot be called from a running event loop (which started by FastAPI).

PR #170, opened 2021-01-31
Feature: currency rates

advaa123:
Added daily currency rates feature.


app/internal/currency.py
```diff
@@ -0,0 +1,14 @@
+import requests
+
+
+def is_valid_currency_api_response(date: str) -> bool:
+    """Checks for valid response from the currency-API request
+    (returns bool)"""
+
+    try:
+        response = requests.get(f'https://api.exchangeratesapi.io/{date}')
```
yammesicka, line 9: Prefer moving this code to the user end, to prevent high load on the server side :)

app/templates/currency.html
```diff
@@ -0,0 +1,80 @@
...
+                    <ul class="list-group list-group-flush" id="currency">
+                        <!-- JS currency content -->
+                    </ul>
+                </div>
+            </div>
+        </div>
+    </div>
+    {% endif %}
+    <script>
+        function getCurrency(base_currency) {
```
yammesicka, line 31: Put this code in a separate JS file to separate the logic from the view

app/templates/currency.html
```diff
@@ -0,0 +1,80 @@
...
+            var x = document.getElementById("currency");
+            function Get(yourUrl) {
+                var Httpreq = new XMLHttpRequest(); // a new request
```
yammesicka, line 34: `httpRequest =`

app/templates/currency.html
```diff
@@ -0,0 +1,80 @@
...
+                        <!-- JS currency content -->
+                    </ul>
+                </div>
+            </div>
+        </div>
+    </div>
+    {% endif %}
+    <script>
+        function getCurrency(base_currency) {
+            var x = document.getElementById("currency");
```
yammesicka, line 32: Prefer better names for the variables

yammesicka: Also, please use const/var

app/templates/currency.html
```diff
@@ -0,0 +1,80 @@
...
+                return false;
+            }
+
+            var text = `<li class="list-group-item title_size_small"><strong>BASE: `
+                + base_currency + `</strong></li>`;
+
+            for (key in data.rates) {
+                if (key !== base_currency) {
+                    text += `<li class="list-group-item title_size_small">`
+                        + `<a href="javascript:void(0)" onclick='getCurrency("` + key + `");return fasle;'>`
```
yammesicka, line 52: Use addEventListener instead `onclick`

app/templates/currency.html
```diff
@@ -0,0 +1,80 @@
...
+                return Httpreq.responseText;
+            }
+
+            var data = JSON.parse(Get('https://api.exchangeratesapi.io/{{ date|safe }}?base=' + base_currency));
+            if ('error' in data) {
+                x.innerHTML = "";
+                return false;
+            }
+
+            var text = `<li class="list-group-item title_size_small"><strong>BASE: `
```
yammesicka, line 46: Use `document.createElement` instead

tests/test_currency.py
```diff
@@ -0,0 +1,33 @@
...
+    assert is_valid_currency_api_response('')
+
+
+def test_response_error(requests_mock_error):
+    assert not is_valid_currency_api_response('')
+
+
+def test_router_good(client, requests_mock_good):
+    resp = client.get(CURRENCY)
+    assert resp.status_code == status.HTTP_200_OK
```
yammesicka, line 22: Prefer `.ok`

yammesicka (changes requested):
That's a great start! Please take a look at my insights :)

app/static/currency.js
```diff
@@ -0,0 +1,54 @@
+function getCurrency(baseCurrency, dateToday) {
+    const showCurrencyElement = document.getElementById("showCurrency")
+    const currencyViewElement = document.getElementById("currencyView");
+    const currencyButtonElement = document.getElementById("currencyButton");
+    currencyViewElement.innerHTML = "";
+
+    function getResponse(yourUrl) {
+        var httpRequest = new XMLHttpRequest();
```
yammesicka, line 8: Prefer using `let`

app/static/currency.js
```diff
@@ -0,0 +1,54 @@
...
+        httpRequest.open("GET", yourUrl, false);
+        httpRequest.send(null);
+        return httpRequest.responseText;
```
yammesicka, line 11: Prefer managing the request asynchronously :)

app/static/currency.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    const data = JSON.parse(getResponse('https://api.exchangeratesapi.io/' + dateToday + '?base=' + baseCurrency));
```
yammesicka, line 14: Maybe we should put the address as `const`?

app/static/currency.js
```diff
@@ -0,0 +1,54 @@
...
+    if (!('rates' in data)) {
+        return false;
+    }
+
+    if (showCurrencyElement.style.display == 'none') {
```
yammesicka, line 19: Prefer using `data-` attributes instead of checking what is the element's style.

Mixing logic and styling is usually not a good idea :)

app/static/currency.js
```diff
@@ -0,0 +1,54 @@
...
+        showCurrencyElement.style.display = 'block';
+    }
+
+    var listItem = document.createElement("li");
```
yammesicka, line 23: const/let

app/static/currency.js
```diff
@@ -0,0 +1,54 @@
...
+    listItem.className = "list-group-item title_size_small";
+    const baseItem = document.createElement("strong");
+    baseItem.innerHTML = "BASE: " + baseCurrency;
+    listItem.appendChild(baseItem);
+    currencyViewElement.appendChild(listItem);
+
+    for (key in data.rates) {
```
yammesicka, line 30: Separate to functions :)

yammesicka (commented):
Awesome job so far! Let's improve the JS a little bit

app/static/currency.js
advaa123: I don’t know anything about JS so this has been really helpful, thank you for the feedback ❤️

app/templates/currency.html
```diff
@@ -0,0 +1,52 @@
...
+        document.getElementById("currencyView").addEventListener("click", function (e) {
+            if (e.target && e.target.nodeName == "A") {
+                getCurrency(e.target.id, "{{ date | safe }}");
+            }
+        }, false);
+    </script>
+    <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.0.0-beta1/dist/js/bootstrap.bundle.min.js"
+        integrity="sha384-ygbV9kiqUc6oa4msXn9868pTtWMgiQaeYH7/t7LECLbyPA2x65Kgf80OJFdroafW"
+        crossorigin="anonymous"></script>
+    <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js"
```
yammesicka, line 41: Do you really need the jQuery over here?

app/static/currency.js
```diff
@@ -0,0 +1,71 @@
+function getCurrency(baseCurrency, dateToday) {
```
yammesicka, line 1: You should be proud of yourself for this code. Very well done

yammesicka (changes requested):

app/templates/currency.html
advaa123: It’s for the bootstrap button collapse, doesn’t work without it 🙏🏼❤️

app/static/currency.js
advaa123: I learned so much, thank you so much for the feedback!

app/templates/currency.html
yammesicka: You've used both Bootstrap 4 and Bootstrap 5 ^^"

Bootstrap 5 doesn't require jQuery:
https://getbootstrap.com/docs/5.0/components/collapse/

yammesicka (changes requested):

yammesicka (approved):

PR #195, opened 2021-02-03
Feature/login

kobyfogel:
This is my version for  users management system: register, login, and current logged-in user.
This version doesn't use fast-api_users library.
My comment to 1 CR was not answered, so it's not 'resolved' yet.
Instructions for using security dependencies are given in README.md file.


app/routers/login.py
```diff
@@ -0,0 +1,89 @@
...
+@router.get('/user')
+async def user_route(request: Request, current_user: User = Depends(current_user)):
+    if current_user:
+        print(current_user.username)
+    else:
+        print(None)
+    return templates.TemplateResponse("home.html", {
+        "request": request,
+        "message": "user.username"
+    })
```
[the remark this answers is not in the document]
kobyfogel: There are two different versions for login management. This version is one of them. It is not clear yet which one of them will be used in production. This PR is just for CR for now. Please read the description.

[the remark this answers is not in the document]
Gonzom: I did read the description which said:

> Included few functions/routs not for production, just for testing functionality. Those are marked as "# Not for production".

Which is why I asked about the testing. You did not state that they are temporarily in the code until you decide which version you'll use in the future, nor did you document it in the code itself.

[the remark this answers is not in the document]
kobyfogel: will add it to description. thanks

app/routers/login.py
```diff
@@ -0,0 +1,92 @@
...
+    response = RedirectResponse(next, status_code=HTTP_302_FOUND)
+    response.set_cookie(
+        "Authorization",
+        value=jwt_token,
+        httponly=True,
+    )
+    return response
+
+
+# Not for production
```
ellenc345, line 63: if you will push those functions to the repository the routes will be accecible to production? if yes mayby they souled be protected?

kobyfogel: there is an explanation about these functions in the PR description.
if it doesn't answer your question, so i didn't really understand what you are asking..

.env
```diff
@@ -0,0 +1 @@
+DATABASE_CONNECTION_STRING = "sqlite:///calendar.db"
```
yammesicka, line 1: Remove the .env file

app/config.py.example
```diff
@@ -46,6 +46,12 @@ email_conf = ConnectionConfig(
     USE_CREDENTIALS=True,
 )
 
+# register
+DATABASE_CONNECTION_STRING = "sqlite:///calendar.db"
+
+# security
+JWT_KEY = "c06857128217c661de43d746ecdbe9f1dbc3b3685957fab64512456f639b1881"
```
yammesicka, line 53: Add request to change this token in the `README.md` (add one-liner python code to generate such code using `secrets`. make sure it's at least 32 bytes long to prevent brute-forcing, as we use HS256)

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,124 @@
...
+from fastapi import Depends, HTTPException
+from fastapi.security import OAuth2PasswordBearer
+import jwt
+from starlette.requests import Request
+from starlette.responses import RedirectResponse
+from starlette.status import HTTP_401_UNAUTHORIZED
+from .schema import LoginUser
+
+
+JWT_MIN_EXP = 3
```
yammesicka, line 17: Move to config :)

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,124 @@
...
+def get_hashed_password(password) -> str:
+    '''Hashing user password'''
+    return pwd_context.hash(password)
+
+
+def verify_password(plain_password, hashed_password) -> bool:
+    '''Verifying password and hashed password are equal'''
+    try:
+        return pwd_context.verify(plain_password, hashed_password)
+    except Exception:
```
yammesicka, line 37: Please do not cache Exception - be more specific

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+JWT_MIN_EXP = 3
+pwd_context = CryptContext(schemes=["bcrypt"])
+oauth_schema = OAuth2PasswordBearer(tokenUrl="/login")
+
+
+async def get_db_user_by_username(username: str) -> User:
+    '''Checking database for user by username field'''
+    session = SessionLocal()
```
yammesicka, line 24: Use `Depends(get_db)` instead

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,124 @@
...
+            return db_user
+        else:
+            raise HTTPException(
+                status_code=HTTP_401_UNAUTHORIZED,
+                headers=path,
+                detail="Your token is incorrect. Please log in again")
+    except Exception as e:
+        if logged_in:
+            return None
+        if type(e).__name__ == 'ExpiredSignatureError':
```
yammesicka, line 87: Please use the `except` clause to check which type of Exception you cought

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,124 @@
...
+    Check whether JWT token is correct. Returns User object if yes.
+    Returns None or raises HTTPException,
+    depanding which depandency activated this function.
+    '''
+    try:
+        jwt_payload = jwt.decode(
+            token, JWT_KEY, algorithms=JWT_ALGORITHM)
+        jwt_username = jwt_payload.get("sub")
+        jwt_hashed_password = jwt_payload.get("hashed_password")
+        db_user = await get_db_user_by_username(username=jwt_username)
```
yammesicka, line 76: You usually don't want to talk with the database when checking the validity of the JWT token - that's one of the benefits of JWT.

app/internal/user.py
```diff
@@ -0,0 +1,48 @@
+from app.database import models, schemas
+from sqlalchemy.orm import Session
+from app.internal.security.ouath2 import pwd_context
+
+# pwd_context = CryptContext(schemes=["bcrypt"])
+
+
+def get_by_id(db: Session, user_id: int) -> models.User:
```
yammesicka, line 8: Prefer Depends

app/internal/user.py
```diff
@@ -0,0 +1,48 @@
+from app.database import models, schemas
+from sqlalchemy.orm import Session
+from app.internal.security.ouath2 import pwd_context
+
+# pwd_context = CryptContext(schemes=["bcrypt"])
```
yammesicka, line 5: Remove if not necessary

app/internal/user.py
```diff
@@ -0,0 +1,48 @@
...
+    }
+    db_user = models.User(**user_details)
+    db.add(db_user)
+    db.commit()
+    db.refresh(db_user)
+    return db_user
+
+
+def delete_by_mail(db: Session, email: str) -> None:
+    '''deletes a user from database by unique email'''
```
yammesicka, line 45: Prefer using `"""`

app/routers/register.py
```diff
@@ -0,0 +1,88 @@
...
+    return templates.TemplateResponse("register.html", {
+        "request": request,
+        "errors": None
+    })
+
+
+@router.post("/register")
+async def register(
+                request: Request, db:
+                Session = Depends(get_db)) -> templates:
```
yammesicka, line 45: Prefer not to break lines line that :)

app/routers/register.py
```diff
@@ -0,0 +1,88 @@
...
+    form = await request.form()
+    form_dict = dict(form)
+    try:
+        # creating pydantic schema object out of form data
+
+        new_user = UserCreate(**form_dict)
+    except ValidationError as e:
+        # if pydantic validations fails, rendering errors to register.html
+
+        errors = {error['loc'][0]: " ".join((
```
yammesicka, line 56: Hard to read, please reformat

app/routers/register.py
```diff
@@ -0,0 +1,88 @@
...
+            "request": request,
+            "errors": errors,
+            "form_values": form_dict})
+    try:
+        # attempt creating User Model object, and saving to database
+
+        user.create(db=db, user=new_user)
+
+        '''
+        if creating User Model objects fails due to registered unique details -
```
yammesicka, line 69: Use `#` for comments

yammesicka (changes requested):
Awesome job! :)

Please prepare this commit for production.

app/config.py.example
kobyfogel: This is how i generated the current key. But i don't really understand what do you mean. To add that python 1 line code to the README.MD file? But keep it in config so that the code will still work until final production?

app/internal/security/ouath2.py
kobyfogel: https://github.com/tiangolo/fastapi/issues/1881
I already tried to do it when i built the function. To the best of my knowledge, you can't use Depends in a function,  but only when you define a route. I think the reason why is that Depends needs fastapi instance in order to retrieve the expected return, and not a Depend object.
Any thoughts?  or can i leave it as it is?
image

app/internal/security/ouath2.py
kobyfogel: Validating the token is the lesser reason for that. I can do it by comparing to to the SECRET_KEY, and just by successfully decoding the token.
The main reason for that is: Already using the authentication function to return the User instance to the protected route. This way everybody can use it for other functionalities, without the need to query in each route. The dependency itself,  is already returning the User object with all its parameters.
Do you prefer that i will change that?

app/internal/user.py
kobyfogel: Same as above, regarding the ability to use Depends in a regular function and not in a route.

app/config.py.example
yammesicka: You should add an instruction in the README.md about how you generate this token and put it in the `config.py` file.

Every user have to generate his own key, it's not safe to have the publicly advertised key.

app/internal/security/ouath2.py
yammesicka: From your like it seems it isn't an issue, it's just we shouldn't use `Depends`. Sababa.
I still don't really like the idea of creating a new session.
Can you pass the database session as a parameter?

yammesicka (changes requested):

app/routers/register.py
yammesicka: Please don't ever mark a comment as a resolved one unless you've addressed it.

I've asked it multiple times, both in the group and in the [forum](https://forums.pythonic.guru/t/code-review-peer-review/8862).

app/routers/register.py
yammesicka: Please don't ever mark a comment as a resolved one unless you've addressed it.

I've asked it multiple times, both in the group and in the [forum](https://forums.pythonic.guru/t/code-review-peer-review/8862).

app/internal/user.py
yammesicka: Please don't ever mark a comment as a resolved one unless you've addressed it.

I've asked it multiple times, both in the group and in the [forum](https://forums.pythonic.guru/t/code-review-peer-review/8862).

app/config.py.example
```diff
@@ -48,6 +48,13 @@ email_conf = ConnectionConfig(
     USE_CREDENTIALS=True,
 )
 
+# register
+DATABASE_CONNECTION_STRING = "sqlite:///calendar.db"
```
yammesicka, line 52: You already have `DEVELOPMENT_DATABASE_STRING`

README.md
```diff
@@ -44,6 +44,22 @@ uvicorn app.main:app --reload
 python -m pytest --cov-report term-missing --cov=app tests
 ```
 
+### Generating  personal JWT Secret Key
+'''shell
```
yammesicka, line 48: Prefer supplying a one-liner they can run and it would just work. Something like:

```
python -c "
import secrets; from pathlib import Path;
f = Path('app/config.py');
f.write_text(f.read_text().replace('JWT_KEY_PLACEHOLDER', secrets.token_hex(32), 1));
"
```

Put this in the Linux & Windows instruction, and change in the config the line to `JWT_KEY = "JWT_KEY_PLACEHOLDER"`

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    A dependency function.
+    Returns logged in User object if exists.
+    None if not.
+    """
+    if 'Authorization' in request.cookies:
+        jwt = request.cookies['Authorization']
+    else:
+        return None
+    user = await check_jwt_token(db, jwt)
```
yammesicka, line 32: You can just `return await` :)

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,33 @@
+from typing import Union
+
+from app.database.database import get_db
+from app.internal.security.ouath2 import (
+    Depends, Session, check_jwt_token, get_cookie)
+from app.internal.security.schema import CurrentUser
+from starlette.requests import Request
+
+
+async def current_user_required(
```
yammesicka, line 10: `is_logged_in` maybe?

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,127 @@
...
+oauth_schema = OAuth2PasswordBearer(tokenUrl="/login")
+
+
+async def get_db_user_by_username(db: Session, username: str) -> User:
+    """Checking database for user by username field"""
+    user = db.query(User).filter_by(username=username).first()
+    return user
+
+
+def get_hashed_password(password) -> str:
```
yammesicka, line 28: Add type annotations

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,127 @@
...
+    """Hashing user password"""
+    return pwd_context.hash(password)
+
+
+def verify_password(plain_password, hashed_password) -> bool:
```
yammesicka, line 33: Add type annotations

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def create_jwt_token(
+        user: LoginUser, JWT_MIN_EXP: int = JWT_MIN_EXP,
+        JWT_KEY: str = JWT_KEY) -> str:
+    """Creating jwt-token out of user unique data"""
+    expiration = datetime.utcnow() + timedelta(minutes=JWT_MIN_EXP)
+    jwt_payload = {
+        "sub": user.username,
+        "hashed_password": user.password,
```
yammesicka, line 56: Is there a good reason to store the user's password in the token?

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,127 @@
...
+    Check whether JWT token is correct. Returns User object if yes.
+    Returns None or raises HTTPException,
+    depanding which depandency activated this function.
+    """
+    try:
+        jwt_payload = jwt.decode(
+            token, JWT_KEY, algorithms=JWT_ALGORITHM)
+        jwt_username = jwt_payload.get("sub")
+        jwt_hashed_password = jwt_payload.get("hashed_password")
+        db_user = await get_db_user_by_username(db, username=jwt_username)
```
yammesicka, line 77: Why? Please read about JWT tokens - one of the main benefits is the performance boost gain from not talking with the DB

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,127 @@
...
+async def my_exception_handler(
+        request: Request,
+        exc: HTTP_401_UNAUTHORIZED) -> RedirectResponse:
+    """
+    Whenever HTTP_401_UNAUTHORIZED is raised,
+    redirecting to login route, with original requested url,
+    and details for why original request failed.
+    """
+    paramas = f"?next={exc.headers}&message={exc.detail}"
+    url = f"/login{paramas}"
```
yammesicka, line 124: Please use url_for

app/main.py
```diff
@@ -29,10 +34,23 @@ def create_tables(engine, psql_environment):
...
 
+app.include_router(security_testing_routes.router)
+app.include_router(profile.router)
+app.include_router(event.router)
+app.include_router(agenda.router)
+app.include_router(register.router)
+app.include_router(email.router)
+app.include_router(invitation.router)
+app.include_router(login.router)
+app.include_router(logout.router)
```
yammesicka, line 45: Why have you re-added them?

app/main.py
```diff
@@ -1,16 +1,21 @@
...
     agenda, calendar, categories, currency, dayview, email,
-    event, invitation, profile, search, telegram, whatsapp
+    event, invitation, login, logout, profile, register,
+    search, telegram, whatsapp
 )
+from fastapi import Depends, FastAPI, Request
+from fastapi.staticfiles import StaticFiles
+from starlette.status import HTTP_401_UNAUTHORIZED
+from sqlalchemy.orm import Session
+from tests import security_testing_routes
```
yammesicka, line 18: How does it belong here?

app/routers/logout.py
```diff
@@ -0,0 +1,17 @@
...
+    prefix="",
+    tags=["/logout"],
+    responses={404: {"description": "Not found"}},
+)
+
+
+@router.get('/logout')
+async def logout(request: Request):
+    response = RedirectResponse(url="/login", status_code=HTTP_302_FOUND)
+    response.delete_cookie("Authorization")
```
yammesicka, line 16: What if it doesn't exists?

app/routers/register.py
yammesicka: Why you keep on closing it? That's a bit rude.

yammesicka (changes requested):

app/routers/register.py
kobyfogel: I thought your comment was on my annotations, which i addressed.
A miss understanding.

app/routers/register.py
kobyfogel: but i did changed that..

app/internal/security/ouath2.py
kobyfogel: Validating the token is the lesser reason for that. I can do it by comparing to to the SECRET_KEY, and just by successfully decoding the token.
The main reason for that is: Already using the authentication function to return the User instance to the protected route. This way everybody can use it for other functionalities, without the need to query in each route. The dependency itself, is already returning the pydantic object with all the Base.User  object parameters, just without the password.
I found it extremely useful in other functionalities.
Do you prefer that i will change that?

app/routers/logout.py
kobyfogel: Nothing happens. can be done this way with delete_cookie function. i already checked it before.

app/internal/security/ouath2.py
kobyfogel: https://www.starlette.io/routing/
For the reason i am not in a route, but in a regular function, i don't have the router instance over there.
Using request.url_for is possible, but because i want to pass query parameters, it is much more complicated.
In flask it's much more simple..
After reading allot, i really think that the way i wrote it is much more simple, then trying to pass parameters the same way the documentation suggests.
Can i leave this one as it is?

kobyfogel: I have tried to do it. url_for uses router, which i don't have here and i have to import it. i have tried to do so, but i got stuck with circular importing that i could not solve. I also tried to move this function to login.py, where i have the router, but still got circular importing, this time with testing. i was unable to resolve it, and i hope we can just leave this one like that.. :)

kobyfogel:
If i clicked on 'resoved' on an issue, without addressing it, it was just by mistake. In few of the cases, i just thought that the CR is for another line of code, and i thought i did resolve it.

app/routers/logout.py
yammesicka: Cool

README.md
```diff
@@ -37,13 +39,30 @@ source venv/bin/activate
...
-uvicorn app.main:app --reload
+# Rendering JWT_KEY:
+python -c "import secrets; from pathlib import Path; f = Path('app/config.py'); f.write_text(f.read_text().replace('JWT_KEY_PLACEHOLDER', secrets.token_hex(32), 1));"
 ```
 ### Running tests
 ```shell
 python -m pytest --cov-report term-missing --cov=app tests
 ```
 
+### Generating  personal JWT Secret Key
```
yammesicka, line 50: Great job editing the `README.md` file. How about removing the two following paragraphs? It isn't a developer guide, just an initial readme to set the project up :)

app/internal/security/ouath2.py
yammesicka: Sorry, but I still unable to understand why.

Why not having one method to check the username and the password against the database (which happens only if the user asks to log-in), and another method to check if the user is authenticated (have a valid JWT token)?

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,128 @@
...
+def verify_password(plain_password: str, hashed_password: str) -> bool:
+    """Verifying password and hashed password are equal"""
+    return pwd_context.verify(plain_password, hashed_password)
+
+
+async def authenticate_user(
+        db: Session, user: LoginUser) -> Union[LoginUser, bool]:
+    """Verifying user is in database and password is correct"""
+    db_user = await get_db_user_by_username(db=db, username=user.username)
+    if db_user:
```
yammesicka, line 42: Use `and` instead two different `if` statements

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def get_hashed_password(password: str) -> str:
+    """Hashing user password"""
+    return pwd_context.hash(password)
+
+
+def verify_password(plain_password: str, hashed_password: str) -> bool:
+    """Verifying password and hashed password are equal"""
+    return pwd_context.verify(plain_password, hashed_password)
```
yammesicka, line 35: Do we have a salt for each user?

app/internal/security/ouath2.py
yammesicka: :+1: Got you

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    if 'Authorization' in request.cookies:
+        return request.cookies['Authorization']
+    raise HTTPException(
+        status_code=HTTP_401_UNAUTHORIZED,
+        headers=request.url.path,
+        detail="Please log in to enter this page")
+
+
+async def my_exception_handler(
```
yammesicka, line 116: What do you mean by `my_`? Is that `auth_exception_handler`? `login_exception_handler`?

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,128 @@
...
+    depanding which depandency activated this function.
+    """
+    try:
+        jwt_payload = jwt.decode(
+            token, JWT_KEY, algorithms=JWT_ALGORITHM)
+        jwt_username = jwt_payload.get("sub")
+        jwt_user_id = jwt_payload.get("user_id")
+        db_user = await get_db_user_by_username(db, username=jwt_username)
+        if db_user and db_user.id == jwt_user_id:
+            return CurrentUser(**db_user.__dict__)
```
yammesicka, line 80: What is the difference between `CurrentUser` and `LoginUser`?

yammesicka (changes requested):
Awesome progress - keep up the good job :)

app/internal/security/ouath2.py
kobyfogel: LoginUser is a pydantic object for internal actions management  of the security system.
CurrentUser is the Object that will be returned to the routes from the dependencies functions, after verifying JWT token.
I'm thinking maybe to re-change it to a Base Model object, as the code is keep on changing by other. And this object has to be maintained in order to keep it's functionality.

yammesicka: What are the downsides of having only one of them? Currently having two different classes for two similar objects is a bit confusing from the developer side

yammesicka (changes requested):

app/internal/security/ouath2.py
kobyfogel: If by 'them' you mean LoginUser and CurrentUser, so they are not the same at all. LoginUser is created from the login form, and verifies that required information was given by the user. It is essential for this action. As CurrentUser supposed to return all the User information to the routes, after jwt verification.
Maybe it will be better to delete CurrentUser, and just return the User Base Object? It's a db object, no confusions there.

yammesicka: I think the major source of confusion is in the object's name. Maybe we should find a better name which describes it better and which highlights the differences between them.

kobyfogel: if it ok by you, i will just return db_user. otherwise, every change in User base model will have to be updated in CurrentUser, in order for it to return the new column.

app/internal/security/ouath2.py
kobyfogel: https://passlib.readthedocs.io/en/stable/lib/passlib.hash.bcrypt.html
auto generated by the 'bcrypt' scheme i use. as well as 12 rounds

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+async def is_logged_in(
+    request: Request, db: Session = Depends(get_db),
+        jwt: str = Depends(get_cookie)) -> bool:
+    """
+    A dependency function protecting routes for only logged in user
+    """
+    await check_jwt_token(db, jwt)
+    return True
```
yammesicka, line 38: Will this function always return `True`?

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+from app.dependencies import get_db
+from app.database.models import User
+from app.internal.security.ouath2 import (
+    Depends, Session, check_jwt_token,
+    get_cookie, get_db_user_by_username,
+    HTTP_401_UNAUTHORIZED, HTTPException)
+from starlette.requests import Request
+
+
+async def validate_user(
```
yammesicka, line 12: Hasn't this function should have been called `get_user`?

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+    request: Request, db: Session = Depends(get_db),
+        jwt: str = Depends(get_cookie)) -> bool:
+    """
+    A dependency function protecting routes for only logged in user
+    """
+    await check_jwt_token(db, jwt)
+    return True
+
+
+async def is_authenticated(
```
yammesicka, line 41: `is_` is a prefix of something that returns either `True` or `False`.

You should either change the function or the name of it :)

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
+from typing import Union
```
yammesicka, line 1: I believe some problems in this CR could be resolved with better naming.

We should also rethink a bit about the design: there are some functions that it's hard to tell why they're different one from another.

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+    A dependency function protecting routes for only logged in user.
+    Returns logged in User object.
+    """
+    username, user_id = await check_jwt_token(db, jwt)
+    return await validate_user(db, user_id, username, request.url.path)
+
+
+async def current_user(
+    request: Request,
+        db: Session = Depends(get_db)) -> Union[User, bool]:
```
yammesicka, line 55: The type annotation here is wrong

app/internal/security/ouath2.py
yammesicka: Awesome!

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,117 @@
...
+from starlette.responses import RedirectResponse
+from starlette.status import HTTP_401_UNAUTHORIZED
+from .schema import LoginUser
+
+
+pwd_context = CryptContext(schemes=["bcrypt"])
+oauth_schema = OAuth2PasswordBearer(tokenUrl="/login")
+
+
+async def get_db_user_by_username(db: Session, username: str) -> User:
```
yammesicka, line 22: I believe this function should be inside the `User` model, or on `/app/internal/users.py`, and named `by_username`.

Keep it simple and think about other developers who want to use it :)

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,117 @@
...
+    db_user = await get_db_user_by_username(db=db, username=user.username)
+    if db_user and verify_password(user.password, db_user.password):
+        return LoginUser(
+            user_id=db_user.id,
+            username=user.username, password=db_user.password)
+    return False
+
+
+def create_jwt_token(
+        user: LoginUser, JWT_MIN_EXP: int = JWT_MIN_EXP,
```
yammesicka, line 50: Don't call parameters with ALL_CAPS. you can use `jwt_expiration_time` and `jwt_key`

app/internal/security/ouath2.py
```diff
@@ -0,0 +1,117 @@
...
+            headers=path,
+            detail="Your token has expired. Please log in again")
+    except jwt.DecodeError:
+        raise HTTPException(
+            status_code=HTTP_401_UNAUTHORIZED,
+            headers=path,
+            detail="Your token is incorrect. Please log in again")
+
+
+async def get_cookie(request: Request) -> str:
```
yammesicka, line 92: The name should be `get_authorization_cookie` (this function's purpose is not "getting any cookie")

app/internal/user.py
```diff
@@ -0,0 +1,46 @@
...
+from app.internal.security.ouath2 import get_hashed_password
+from sqlalchemy.orm import Session
+
+
+def get_by_id(db: Session, user_id: int) -> models.User:
+    """query database for a user by unique id"""
+    return db.query(models.User).filter(models.User.id == user_id).first()
+
+
+def get_by_username(db: Session, username: str) -> models.User:
```
yammesicka, line 11: How does this function is different from `get_db_user_by_username`?

app/routers/login.py
```diff
@@ -0,0 +1,72 @@
...
+        "request": request,
+        "message": message,
+        'current_user': current_user
+    })
+
+
+@router.post('/login')
+async def login(
+        request: Request,
+        next: Optional[str] = "/",
```
yammesicka, line 38: We should check `next` is not malicious

app/routers/register.py
```diff
@@ -0,0 +1,81 @@
...
+    prefix="",
+    tags=["register"],
+    responses={404: {"description": "Not found"}},
+)
+
+
+async def render_registration_error(
+        db: Session, new_user: UserCreate) -> dict:
+    """After registering new user fails, defines relevant errors"""
+    db.rollback()
```
yammesicka, line 22: Why should we rollback the DB first?

app/routers/login.py
```diff
@@ -0,0 +1,72 @@
...
+        db: Session = Depends(get_db),
+        existing_jwt: Union[str, bool] = False) -> RedirectResponse:
+    """rendering login route post method."""
+    form = await request.form()
+    form_dict = dict(form)
+    # creating pydantic schema object out of form data
+
+    user = LoginUser(**form_dict)
```
yammesicka, line 46: You should use `schemas.LoginUser`, that will make it much more obvious :)

app/routers/register.py
```diff
@@ -0,0 +1,81 @@
...
+    errors = {}
+    db_user_email = user.get_by_mail(db, email=new_user.email)
```
yammesicka, line 24: Prefer using query with `exists`

app/routers/register.py
```diff
@@ -0,0 +1,81 @@
...
+        new_user = UserCreate(**form_dict)
+    except ValidationError as e:
+        # if pydantic validations fails, rendering errors to register.html
+
+        errors_dict = {error['loc'][0]: error['msg'] for error in e.errors()}
+        errors = {
+            field: " ".join((
+                field.capitalize(), error_message)
+                ) for field, error_message in errors_dict.items()}
+
```
yammesicka, line 62: These lines are still unreadable.

```python
def get_error_messages_by_fields(errors: [type here])-> Dict[str, str]:
    errors_by_fields = {error['loc'][0]: error['msg'] for error in e.errors()}
    return {
        field_name: f"{field.capitalize()} {error_message}"
        for field_name, error_message in e.errors())
    }
```

app/routers/register.py
```diff
@@ -0,0 +1,81 @@
...
+        return templates.TemplateResponse("register.html", {
+            "request": request,
+            "errors": errors,
+            "form_values": form_dict})
+    try:
+        # attempt creating User Model object, and saving to database
```
yammesicka, line 68: Verbose comments in my opinion

app/templates/base.html
```diff
@@ -31,11 +31,13 @@
           <a class="nav-link active" aria-current="page" href="{{ url_for('profile') }}">{{ gettext("Profile") }}</a>
         </li>
         <li class="nav-item">
-          <a class="nav-link" href="#">{{ gettext("Sign in") }}</a>
+            <a class="nav-link" href="{{ url_for('register') }}">{{ gettext("Sign Up") }}</a>
         </li>
         <li class="nav-item">
-          <a class="nav-link" href="#">{{ gettext("Sign up") }}</a>
+            <a class="nav-link" href="{{ url_for('login') }}">{{ gettext("Sign in") }}</a>
```
yammesicka, line 37: In

yammesicka (commented):
Great job! Another review to go :)

app/internal/security/ouath2.py
kobyfogel: I know it does.
The reason i have it there is because i though that it will be 'more corrects' to build the login and authentication systems as a stand-alone system. A code that needs the minimum changes if someone wants to use it in another project.
I will delete it if you prefer, and use the other function.

app/internal/user.py
kobyfogel: I can use this one, but i tried to explain above why i did that.

app/internal/security/dependancies.py
kobyfogel: this function activates check_jwt_token function. That function will raise an exception  if validation fails. so yes..
should i add that to annotations?

app/internal/security/ouath2.py
yammesicka: I do prioritize DRY code in this case :)

app/routers/login.py
kobyfogel: I don't really understand what do you mean. This is the requested url, all i' am trying to do is to 'remember' it, if it is under dependency, and if the user was not logged in when sending the request.
The goal is is simply to return the user directly to the requested route, after logging in.
In this case, next is a parameter  i created in ouath2.auth_exception_handler. I chose to use that name because it is consensus name for this parameter. should i change that?

app/routers/register.py
kobyfogel: There was an issue when IntegrityError is raised by sqlalchemy. 
image
rolling back is the solution i found.

app/internal/security/ouath2.py
yammesicka: Just be sure to name stuff properly. Otherwise it'll be really PITA to maintain :)

app/internal/security/dependancies.py
yammesicka: So catch the exception and return `False`.
The function called `is_logged_in`, the developer who sees it surely expect to get `True`/`False` :)

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+    A dependency function protecting routes for only logged in user.
+    Returns logged in User object.
+    """
+    username, user_id = await check_jwt_token(db, jwt)
+    return await get_user(db, user_id, username, request.url.path)
+
+
+async def optional_logged_in_user(
+    request: Request,
+        db: Session = Depends(get_db)) -> Union[User, type(None)]:
```
yammesicka, line 55: 1. `Optional[User]` (instead of `Union` with `NoneType`)
2. Remove the indentation before the `db:`
3. Prefix `)` in a new line, with comma after the previous line.

Look how the code become much more readable:

```python
async def current_user(
    request: Request, db: Session = Depends(get_db),
) -> Optional[User]:
```

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    A dependency function.
+    Returns logged in User object if exists.
+    None if not.
+    """
```
yammesicka, line 60: Prefer a docstring that matches the guidelines in PEP 257:

```python
    """Return the currently logged in user.

    Return logged in User object if exists, None if not.
    A dependency function.
    """
```

app/routers/login.py
```diff
@@ -0,0 +1,74 @@
...
+    prefix="",
+    tags=["/login"],
+    responses={404: {"description": "Not found"}},
+)
+
+
+@router.get("/login")
+async def login_user_form(
+        request: Request, message: Optional[str] = "",
+        current_user: Union[User, None] = Depends(optional_logged_in_user)
```
yammesicka, line 25: Optional

app/routers/login.py
yammesicka: The `next` can be `http://attacker.com`, for example.

Please read about open redirection attack.

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+    Recives user details from decrypted jwt token,
+    and check them against the database.
+    returns User base model instance if succeeded,
+    raises HTTPException if fails.
+    '''
+    db_user = await User.get_by_username(db, username=username)
+    if db_user and db_user.id == user_id:
+        return db_user
+    else:
+        raise HTTPException(
```
yammesicka, line 25: Prefer not raising it as an http exception - let the route do it.

In these function you need to return a Pythonic response - which should probably be None, in this case.

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+    request: Request, db: Session = Depends(get_db),
+        jwt: str = Depends(get_authorization_cookie)) -> bool:
+    """
+    A dependency function protecting routes for only logged in user
+    """
+    await check_jwt_token(db, jwt)
+    return True
+
+
+async def logged_in_user(
```
yammesicka, line 41: This is not a verb, so it's hard to understand what the function suppose to do.

Should it be `is_logged_in`?
How does it different from the current `is_logged_in`?

app/internal/security/dependancies.py
```diff
@@ -0,0 +1,66 @@
...
+        jwt: str = Depends(get_authorization_cookie)) -> User:
+    """
+    A dependency function protecting routes for only logged in user.
+    Returns logged in User object.
+    """
+    username, user_id = await check_jwt_token(db, jwt)
+    return await get_user(db, user_id, username, request.url.path)
+
+
+async def optional_logged_in_user(
```
yammesicka, line 53: Is that another instance of `get_user`?

app/routers/logout.py
```diff
@@ -0,0 +1,17 @@
...
+
+router = APIRouter(
+    prefix="",
+    tags=["/logout"],
+    responses={404: {"description": "Not found"}},
+)
+
+
+@router.get('/logout')
+async def logout(request: Request):
```
yammesicka, line 14: Can we somehow check the logout request using some CSRF cookie? Currently it will disconnect you if you'll click a [link](https://calendar.pythonic.guru/logout) that I've sent you.

yammesicka: (Don't overthink it if you find it too hard - I do want to finish this ticket ASAP)

app/routers/register.py
```diff
@@ -0,0 +1,99 @@
...
+        errors['username'] = "That username is already taken"
+    if db.query(
+        db.query(models.User).filter(
+            models.User.email == new_user.email).exists()).scalar():
+        errors['email'] = "Email already registered"
+    return errors
+
+
+def get_error_messages_by_fields(errors: [list]) -> Dict[str, str]:
+    '''Getting validation errors by fields from pydantic ValidationError'''
```
yammesicka, line 57: Decide on whether to use `'''` or `"""` (you mixed them inside the file)

app/routers/register.py
```diff
@@ -0,0 +1,99 @@
+from typing import Dict
```
yammesicka, line 1: Separate to 3 paragraphs

app/routers/register.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+router = APIRouter(
+    prefix="",
+    tags=["register"],
+    responses={404: {"description": "Not found"}},
+)
+
+
+async def user_create(
```
yammesicka, line 20: Prefer `create_user`

yammesicka (commented):
Great job and great improvement! Here are some things to improve :)

app/internal/security/dependancies.py
kobyfogel: image
The same goes for both of the comments:
The whole point of these functions, is to protect a certain route from not logged in user be able to reach it.
In any case which that i don't raise an Exception, but returning a  Boolean, that will not do the job. I already checked that, and This is precisely the case with the 'optional_logged_in_user' dependency that i made, which is not protecting the route.
This is why i configured 'auth_exception_handler' the way i did in the first place:
So i can create a dependency to protect a route, and at the same time render a specific message about what failed, and 'remembering' the user's original request

app/internal/security/dependancies.py
kobyfogel: The difference is described in the Type Annotations.
They both protecting a route for only logged in user.
The difference is while is_logged_in returns a Boolean, this one returns the User object itself.
I think its extremely useful having this kind of dependency. Then you don't need to query the database in each route when you want to use the current user's different data.
At first i had just this one, but you did not want the database to be queried when the jwt is checked. So i made two different dependencies that to different actions, thinking about other developers having useful dependencies.

app/internal/security/dependancies.py
kobyfogel: No it's not.
It is designed to be a route dependency.
get_user is a function that helps other dependencies, after the jwt token was extracted from cookie, and was checked.
The way it is done varies with the different dependencies.

app/routers/logout.py
kobyfogel: I started to read about it. Seems like it's a whole topic of security, which is new to me.
I am always eager to learn more but that will take some time to understand in a proper depth.
If we can leave it as it is for now, I promise to try to improve it after i finish my tickets.
If you have any good suggestions for me regarding what on this subject to put my focus on, i will be more then happy to know.

kobyfogel:
I tried to answer the best i can, hopefully it clarifies everything that was not understood.
I also tried to fix what i could, and i hope that will be good enough.

app/routers/logout.py
yammesicka: Sababa :)

app/internal/security/dependancies.py
yammesicka: So how is it different from `get_current_user`?

You must understand you're building an API for other programmers.
All the functions *must* be easily understood from their names only.

Currently it's really hard to tell what this function (or any other function in this file) do from only looking at it's name.
We must find a better naming that will describe it in a way the other programmers would be able to understand what is does.

app/internal/security/dependancies.py
yammesicka: Unfortunately I can't understand what is the difference even after your explanation.

Please try to look at the issue from the perspective of the programmer who uses your system.

He probably wants to have:

1. `is_logged_in`
2. `is_manager`
3. `get_user_by_username` / `_id` / `_email`
4. `login` / `logout`

That's it.

app/routers/register.py
```diff
@@ -0,0 +1,101 @@
+from typing import Dict
+
+from app.internal.security.ouath2 import get_hashed_password
```
yammesicka, line 3: This paragraph should be the 3rd one

yammesicka (changes requested):
Hope we're almost done :)

PR #119, opened 2021-01-22
Feature/import to calendar

Nir-P:
### This feature includes functions that support importing txt, csv, and ics (iCalendar) files to our Calendar.

**txt and csv files must be in the following structure:**

```
Title1, Content1, S_Date1, E_Date1
Title2, Content2, S_Date2, E_Date2
...
TitleN, ContentN, S_DateN, E_DateN
```

the main function "user_click_import" handles the file, checking that the file passes all the rules we set, and if everything is ok the file imported to the calendar and populates the event table.

**some of the rules we set (can be changed through a global variable):**

- Files must be txt, csv, or ics.
- Maximum file size 5 MB.
- txt and csv must be in the structure explained above.
- Date must be with the following mask "MM-DD-YYYY" (for txt and csv).
- The title of the event can't be empty.
- The title length is up to 50 characters.
- The content may be empty.
- The content length is up to 500 characters.
- Maximum events with the same date: 10.
- Events must be within 20 years range before and after the current year.

**UPDATE:** 
- Added support for the start date and end date in the event table.

PR #201, opened 2021-02-04
feat: improve import file to calendar and extend capabilities

Nir-P:
This is an extension for [import to calendar](https://github.com/PythonFreeCourse/calendar/pull/119):

### what's new?

- Support "Location" field when importing txt, csv or ics files 
  files must be in the following structure or the old structure (without Location)
  ```
  Title1, Content1, S_Date1, E_Date1, Location1
  Title2, Content2, S_Date2, E_Date2, Location2
  ...
  TitleN, ContentN, S_DateN, E_DateN, LocationN
  ```
- Start and End dates support hours and minutes format

**the new rules we set (can be changed through a global variable):**
-  The location field length is up to 50 characters.
-  The event duration limit.

**Additional tests added:**
- Start date is smaller or equal to the end date.
- The Location field may be empty.
- Checking that new files from the old and new format are working flawlessly.
- And some more here and there :)


app/internal/import_file.py
```diff
@@ -45,26 +57,54 @@ def is_file_exist(file: str) -> bool:
...
     """
     check if date is valid and in the range according to the rule we have set
     """
     now_year = datetime.datetime.now().year
     if isinstance(date, str):
-        try:
-            check_date = datetime.datetime.strptime(date, DATE_FORMAT)
-        except ValueError:
+        check_date = change_string_to_date(date)
+        if check_date is False:
```
yammesicka, line 79: Prefer `== False`

app/internal/import_file.py
```diff
@@ -45,26 +57,54 @@ def is_file_exist(file: str) -> bool:
     return Path(file).is_file()
 
 
+def change_string_to_date(date: str) -> Union[datetime.datetime, bool]:
+    try:
+        if re.search(r":", date):
```
yammesicka, line 62: Prefer `: in date`

app/internal/import_file.py
```diff
@@ -45,26 +57,54 @@ def is_file_exist(file: str) -> bool:
...
+            date1 = datetime.datetime.strptime(date, DATE_FORMAT2)
+        else:
+            date1 = datetime.datetime.strptime(date, DATE_FORMAT)
+    except ValueError:
```
yammesicka, line 66: What is the purpose of the except here?

yammesicka (changes requested):
Great job! Added few of my insights :)

app/internal/import_file.py
Nir-P: if the string is not a valid date `datetime.strptime` will fail to transfer the string to date and we want to catch it.

app/internal/import_file.py
Nir-P: changed it to `not` if this is ok.

app/internal/import_file.py
Nir-P: Done

yammesicka (approved):

PR #220, opened 2021-02-06
feat: add alembic configuration

ellenc345:
I've ran some tests to add and rollback a column. seems to work fine.


ellenc345:
review?

alembic.ini
```diff
@@ -0,0 +1,86 @@
+# A generic, single database configuration.
+
+[alembic]
+# path to migration scripts
+# script_location = alembic
+script_location = alembic
+
+# template used to generate migration files
+# file_template = %%(rev)s_%%(slug)s
```
yammesicka: Maybe we should do it by date?

alembic.ini
```diff
@@ -0,0 +1,86 @@
...
+# version location specification; this defaults
+# to alembic/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+# version_locations = %(here)s/bar %(here)s/bat alembic/versions
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+sqlalchemy.url = driver://user:pass@localhost/dbname
```
yammesicka: We should find a way to take it from the config.py file

alembic.ini
```diff
@@ -0,0 +1,86 @@
...
+# on newly generated revision scripts.  See the documentation for further
+# detail and examples
+
+# format using "black" - use the console_scripts runner, against the "black" entrypoint
+# hooks=black
+# black.type=console_scripts
+# black.entrypoint=black
+# black.options=-l 79
+
+# Logging configuration
```
yammesicka, line 53: Please find a way to connect the alembic to our `logger` object :)

alembic.ini
```diff
@@ -0,0 +1,86 @@
+# A generic, single database configuration.
```
yammesicka, line 1: Please move all the alembic directory structure into `app/database/alembic`

alembic/env.py
```diff
@@ -0,0 +1,81 @@
+from app.database.database import Base
```
yammesicka:
```
from logging.config import fileConfig
import os

from alembic import context
from sqlalchemy import create_engine

from app import config as app_config
from app.database.database import Base
```

alembic/env.py
```diff
@@ -0,0 +1,81 @@
...
+    In this scenario we need to create an Engine
+    and associate a connection with the context.
+
+    """
+
+    connectable = create_engine(SQLALCHEMY_DATABASE_URL)
+
+    with connectable.connect() as connection:
+        context.configure(
+            connection=connection, target_metadata=target_metadata
```
yammesicka: Consider adding a trailing comma

yammesicka (commented):
Awesome job! Added few of my insights :)

alembic.ini
ellenc345: it is overrides in the env.py, anyway, i commented it out

alembic.ini
yammesicka: Not resolved

alembic.ini
yammesicka: If possible, move the configurations to `config.py.example`

yammesicka (changes requested):

alembic.ini
ellenc345: look at env.py 
its pulling its default logging and uses it only if there is no logger assigned.
`fileConfig(config.config_file_name, disable_existing_loggers=False)`

if i replace entirly the logger, there will be no logs when runing from console.

alembic.ini
ellenc345: Can you please alborate why do that?
this file is runing when you call the alembic into action. this is the standart config that is well documented and it is clear to maintain it. if you whant to override things you can alwayys do it in alembic/env.py file.
but as I see it i wouldn't mix that config with the rest of the custom code configuration.
Ill be glad to hear your opinion about that.

alembic.ini
yammesicka: Cool, gotcha

alembic.ini
yammesicka: I'd like to avoid making the user edit multiple configuration files, but now I see that the user shouldn't really touch anything here, so I guess it's cool

yammesicka (approved):

PR #222, opened 2021-02-06
add: world_clock.py and test_world_clock.py

nir9696:


tests/test_world_clock.py
```diff
@@ -0,0 +1,215 @@
+from app.routers.world_clock import generate_possible_timezone_path,\
```
yammesicka: Prefer splitting using parenthesis (apply this rule also for the rest of the code)

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+            datetime.strptime(current_time_in_place, '%H:%M:%S')
+        delta_in_hours = \
+            int((time_now - current_time_in_place).total_seconds()
+                / TOTAL_SEC_IN_HOUR)
+        attendance_time = (input_time + timedelta(hours=delta_in_hours)).\
+            strftime('%H:%M:%S')
+        attendance_time = datetime.strptime(attendance_time, "%H:%M:%S")
+        part_of_day, feedback = \
+            get_part_of_day_and_feedback(attendance_time)
+        return f"{attendance_time.strftime('%H')}:" \
```
yammesicka: Extract to a better named variables

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+    """
+    now = datetime.now()
+    time_now_str = now.strftime("%H:%M:%S")
+    time_now = datetime.strptime(time_now_str, "%H:%M:%S")
+    input_time = datetime.strptime(time_str, "%H:%M:%S")
+    current_time_in_place = get_current_time_in_place(place_name)
+    if current_time_in_place:
+        current_time_in_place = \
+            datetime.strptime(current_time_in_place, '%H:%M:%S')
+        delta_in_hours = \
```
yammesicka: Do it in two steps to prevent hard to read code

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+    for part_of_day, details in PARTS_OF_THE_DAY_FEEDBACK.items():
+        start_time_scope, end_time_scope = details[TIME_SCOPE_INDEX]
+        feedback = details[FEEDBACK_INDEX]
+        if time >= datetime.strptime(start_time_scope, '%H:%M:%S') \
+                and time <= datetime.strptime(end_time_scope, '%H:%M:%S'):
+            return part_of_day, feedback
+
+
+def meeting_possibility_feedback(time_str: str, place_name: str) \
+        -> Tuple[str, str, str]:
```
yammesicka: What happens if the `not current_time_in_place`?

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+                                "country-and-continent-codes-list/" \
+                                "country-and-continent-codes-list-csv_json/" \
+                                "data/c218eebbf2f8545f3db9051ac893d69c/" \
+                                "country-and-continent-codes-list-" \
+                                "csv_json.json"
+CITY_TO_COUNTRY_DATA_SET = "https://pkgstore.datahub.io/core/world-cities/" \
+                           "world-cities_json/data/" \
+                           "5b3dd46ad10990bca47b04b4739a02ba/" \
+                           "world-cities_json.json"
+TIMEZONES_COUNTRY_SUBCOUNTRY = {
```
yammesicka: We shouldn't manage our own database of timezones. Please use `pytz.all_timezones` or something familiar :)

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+CONTINENT = 'continent'
+COUNTRY = 'country'
+PLACE = 'place'
+CURRENT_TIME_KEY = 'datetime'
+TIMEZONE_PARTS_MAP = {CONTINENT: 0, PLACE: -1}
+TIME_SCOPE_INDEX = 0
+FEEDBACK_INDEX = 1
+ARB_COUNTRY_TIMEZONE = -1
+PATH_SEPARETOR = '/'
+PARTS_OF_THE_DAY_FEEDBACK = {'Early morning':
```
yammesicka: It would probably be better to use NamedTuple here

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+                             'Late night':
+                             [('00:00:00', '04:59:59'),
+                              'Not possible'], }
+
+
+def normalize_continent_name(continent_name: str) -> Optional[str]:
+    """Normalize a given continent name to the continent name
+       that appears in the timezone list.
+
+
```
yammesicka: Only one line of spacing is required here

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+    Args:
+        continent_name (str): The given continent name.
+
+
+    Returns:
+        str: The normalized continent name.
+    """
+    for continent in CONTINENTS:
+        if continent in continent_name:
```
yammesicka: So make `continent_name` a set

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+
+
+    Args:
+        city_name (str): The given city name.
+
+
+    Returns:
+        str: The suitable country name.
+    """
+    details = requests.get(CITY_TO_COUNTRY_DATA_SET).json()
```
yammesicka: We shouldn't use `requests` on the server side. Why we need to send a request to the server? Prefer using JS to find it

yammesicka (changes requested):
That's a good start, let's have the first improvement cycle :)

app/routers/world_clock.py
nir9696: Can you explain why It would be better? I find the dict more convenient and elegant here... :)

app/routers/world_clock.py
nir9696: but `continent_name` is a string... I'm actually check if `continent` is a substring of `continent_name` (like "America" in "South America").
Have I misunderstood you?

app/routers/world_clock.py
nir9696: Actually, I do not quite sure how to integrate js here... Isn't it better to save the data as a file and thus it will not be necessary to send a request to the server?

app/routers/world_clock.py
nir9696: I have looked into several modules in Python, including pytz, and I can not see how they help here. I have invested a lot of time and energy in writing a code that supports large amount(!) of cities and places and not just the "timezone places" themselves.
the "Mini DB" I've created is not just a list of timezones - it maps timezones to countries and subcountries and crucial for the core of my code - Ability to search the equivalent time of a custom city or place.

app/routers/world_clock.py
yammesicka:
```
PARTS_OF_THE_DAY_FEEDBACK = [
     MeetingTime(time='Early Morning', start=..., end=..., desirability=...),
     MeetingTime(...),
}
```

app/routers/world_clock.py
yammesicka: Sorry, my bad. I meant `CONTINENTS` should be a set

app/routers/world_clock.py
yammesicka: This is also a good solution if possible.

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+
+
+    Args:
+        city_name (str): The given city name.
+
+
+    Returns:
+        str: The suitable subcountry name.
+    """
+    details = requests.get(CITY_TO_COUNTRY_DATA_SET).json()
```
yammesicka: This `requests` request should also be on JS or from file.

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+    """Get the data from an API url.
+
+    Args:
+        url (str):  The API url.
+
+    Returns:
+        list: The data.
+    """
+    try:
+        resp = requests.get(url)
```
yammesicka: This one also

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+    Returns:
+        list: The list of possible timezones.
+    """
+    prefix = map_hierarchy[CONTINENT]
+    possibilities = [f"{prefix}/{map_hierarchy[COUNTRY]}",
+                     f"{prefix}/{map_hierarchy[PLACE]}"]
+    return possibilities
+
+
+def get_all_possible_timezone_paths_for_given_place(place_name: str)\
```
yammesicka: Split this function into multiple functions.
Also, prefer not to split lines using `\`

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+        -> Optional[List[str]]:
+    """Get all possible timezone paths for given place.
+
+
```
yammesicka: One blank line is enough here

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+    continent = standardize_continent(continent)
+    if not continent:
+        place_name = standardize_country_or_place(place_name)
+        if place_name in get_timezones_parts('place'):
+            possibilities.append(search_timezone_by_just_place(place_name))
+            return possibilities
+    return possibilities.append(f"{continent}/{place_name}")
+
+
+def get_timezone_path_for_given_place(place_name: str) -> Optional[str]:
```
yammesicka: Split to multiple functions

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+        place_name (str): The given place name.
+
+
+    Returns:
+        str: The timezone path.
+    """
+    possibilities = \
+        get_all_possible_timezone_paths_for_given_place(place_name)
+    timezones_list = get_api_data(TIMEZONES_BASE_URL)
+    if possibilities:
```
yammesicka: Returning always a list from `get_all_possible_timezone_paths_for_given_place` might save you the check here

app/routers/world_clock.py
```diff
@@ -0,0 +1,744 @@
...
+    Args:
+        place_name (str): The given place name.
+
+
+    Returns:
+        str: The timezone path.
+    """
+    possibilities = \
+        get_all_possible_timezone_paths_for_given_place(place_name)
+    timezones_list = get_api_data(TIMEZONES_BASE_URL)
```
yammesicka: Prefer to avoid Hungarian notation: `timezones` is enough

yammesicka (commented):
Added few more comments :)

app/routers/world_clock.py
nir9696: this is the core of my logic and the pivot for the other functions - its for getting the current time around the world. I really don't know what I can do here...
The API requests outside this function can be dealt by treating them as reading a file, but this is not the case here...

nir9696:
Hi :)
Can you please response to the 2 comments I wrote above?

app/routers/world_clock.py
yammesicka: You can move it to JS - there are even example of other people who already did it :)

nir9696: Can you please elaborate? That's exactly the part I do not understand :(
1. Can I leave the call to the API?
2. Will some of the code be in Python and some in JS? How does it actually fit in?
3. Alternatively, does it mean that I have to give up all my code and just use JS? I am really afraid from this because I have invested a lot of time and effort in the code and in addition, I am less familiar with JS.

tests/test_world_clock.py
nir9696: manage to do it except one...

app/routers/world_clock.py
nir9696: I didn't see an extra spacing. I think it is now OK.

app/routers/world_clock.py
nir9696: I didn't see an extra spacing. I think it is now OK.

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
...
+import json
+import requests
+
+from app import config
+
+from datetime import datetime, timedelta
+
+from collections import namedtuple
+
+from typing import Any, Dict, List, Optional, Tuple
```
yammesicka: Reorder according to the project's convention

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
...
+COUNTRY = 'country'
+PLACE = 'place'
+CURRENT_TIME_KEY = 'datetime'
+TIMEZONE_PARTS_MAP = {CONTINENT: 0, PLACE: -1}
+TIME_SCOPE_INDEX = 0
+FEEDBACK_INDEX = 1
+ARB_COUNTRY_TIMEZONE = -1
+PATH_SEPARETOR = '/'
+
+MeetingTime = namedtuple('MeetingTime',
```
yammesicka: Prefer using typing.NamedTuple

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
...
+    Args:
+        part (str):  The given part.
+    Returns:
+        list: a list of the relevant parts of the timezones list.
+    """
+    parse_list = parse_timezones_list()
+    return [timezone[TIMEZONE_PARTS_MAP[part]] for timezone in parse_list]
+
+
+def standardize_country_or_place(place_name: str) -> str:
```
yammesicka: This is `Optional[str]`

app/routers/world_clock.py
yammesicka: You can leave it here if you must, just use another `async` module instead of requests, as `requests` blocks the server and make all the other requests wait until this one resolves.

Use httpx instead

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
...
+        list: The list of possible timezones.
+    """
+    map_hierarchy = {CONTINENT: None, COUNTRY: None, PLACE: None, }
+    possibilities = []
+    map_hierarchy[COUNTRY] = country
+    map_hierarchy[PLACE] = place_name
+    continent = get_continent(country)
+    continent = standardize_continent(continent)
+    map_hierarchy[CONTINENT] = continent
+    print(map_hierarchy)
```
yammesicka: Remove

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
...
+    Args:
+        place_name (str): The given country or place name.
+    Returns:
+        str: The standardized name.
+    """
+    if place_name:
+        return place_name.replace(BLANK, UNDERSCORE).title()
+
+
+def standardize_continent(continent_name: str) -> str:
```
yammesicka: Optional. Please run mypy

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
...
+    """
+    map_hierarchy = {CONTINENT: None, COUNTRY: None, PLACE: None, }
+    possibilities = []
+    map_hierarchy[COUNTRY] = country
+    map_hierarchy[PLACE] = place_name
+    continent = get_continent(country)
+    continent = standardize_continent(continent)
+    map_hierarchy[CONTINENT] = continent
+    print(map_hierarchy)
+    if not continent:
```
yammesicka: Flip the condition to return early

tests/test_world_clock.py
```diff
@@ -0,0 +1,263 @@
+from app.routers.world_clock import \
```
yammesicka: Open the parenthesis one line before, avoid using `\`

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
...
+    Returns:
+        str: The timezone path.
+    """
+    possibilities = (
+        get_all_possible_timezone_paths_for_given_place(place_name))
+    timezones = get_api_data(TIMEZONES_BASE_URL)
+    for possibility in possibilities:
+        if possibility in timezones:
+            return f'{TIMEZONES_BASE_URL}/{possibility}'
+    # NO EXPLICIT SUCH TIMEZONES. GET ARBITRARY TIMEZONE OF COUNTRY
```
yammesicka: Don't use all caps (unless you really want to scream)

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
...
+
+def get_current_time_in_place(place_name: str) -> Optional[str]:
+    """Get the current time in a given place.
+    Args:
+        place_name (str): The given place name.
+    Returns:
+        str: The timezone path.
+    """
+    path = get_timezone_path_for_given_place(place_name)
+    if path:
```
yammesicka: Use guard instead:

```
if not Path:
    return
```

app/routers/world_clock.py
```diff
@@ -0,0 +1,437 @@
+import dateutil.parser
```
yammesicka: Put everything that not directly relates to the `routes` inside the `internal` directory

yammesicka (changes requested):
Hi, I've added further insights.

app/routers/world_clock.py
nir9696: I am having a really hard time with making the code asynchronous and the submission time is already pressing so I am very stressed out.
I researched hours online, asked at StackOverflow, but failed to to put it together into a code that works. 
1. Do I need to change all the functions to "async def"? 
	1.1. Do I have to change the tests to async def as a result? 
2. Should the call to the API be with "await"? why? 
	2.1. If so - for some reason I can not refer to the information obtained from the call to the API in other functions for the reason that "'coroutine' object is not iterable python".

As a beginning, I tried to modify the code of the "get_api_data" function according to the documentation to the following function:
def get_api_data (url: str) -> Optional [List [Any]]:
    try:
        async with httpx.AsyncClient () as client:
             resp = await client.get (url)
           return resp.json ()
    except httpx.HTTPError as errh:
      print ("Http Error:", errh)

But it just does not work ...
Can you guide me?

app/routers/world_clock.py
nir9696: hopefully done :)

app/internal/world_clock.py
```diff
@@ -0,0 +1,438 @@
...
+        int((time_part - time_in_place).total_seconds()
+            / TOTAL_SEC_IN_HOUR))
+    wanted_time_there = ((wanted_time + timedelta(hours=delta_in_hours)).
+                         strftime('%H:%M:%S'))
+    return datetime.strptime(wanted_time_there, "%H:%M:%S")
+
+
+async def meeting_possibility_feedback(
+        wanted_time_here: str, place_name: str) \
+        -> Optional[Tuple[str, str, str]]:
```
yammesicka: Please use precommit hooks to reformat

app/internal/world_clock.py
```diff
@@ -0,0 +1,438 @@
...
+    possibilities = (
+        await get_all_possible_timezone_paths_for_given_place(place_name))
+    timezones = await get_api_data(TIMEZONES_BASE_URL)
+    for possibility in possibilities:
+        if possibility in timezones:
+            return f'{TIMEZONES_BASE_URL}/{possibility}'
+    # No explicit such timezones. Get an arbitrary timezone of country
+    if await standardize_continent(place_name.title()):
+        res = await get_arbitrary_timezone_of_country(place_name)
+        return res
```
yammesicka: Can return directly, without `res =`

app/internal/world_clock.py
```diff
@@ -0,0 +1,438 @@
...
+async def search_timezone_by_just_place(place_name: str) -> Optional[str]:
+    """Search for a timezone in the timezones list by a given place name.
+    Args:
+        place_name (str): The given place name.
+    Returns:
+        str: The suitable timezone.
+    """
+    res = [timezone for timezone in await get_api_data(TIMEZONES_BASE_URL)
+           if place_name in timezone]
+    if len(res):
```
yammesicka: `if res`

app/internal/world_clock.py
```diff
@@ -0,0 +1,438 @@
...
+    Args:
+        place_name (str): The given country or place name.
+    Returns:
+        str: The standardized name.
+    """
+    if place_name:
+        return place_name.replace(BLANK, UNDERSCORE).title()
+
+
+async def standardize_continent(continent_name: str) -> str:
```
yammesicka: Optional?

yammesicka (changes requested):
Great start, just few things before we can merge it

app/internal/world_clock.py
nir9696: Done

app/internal/world_clock.py
```diff
@@ -0,0 +1,512 @@
...
+    Returns:
+        tuple: The part of day description and the feedback.
+    """
+    for part in PARTS_OF_THE_DAY_FEEDBACK:
+        if (
+            time
+            >= datetime.strptime(
+                part.start,
+                "%H:%M:%S",
+            )
```
yammesicka: Export this and the other one to a variable, outside the loop, and create 1 line `if` statement

app/internal/world_clock.py
```diff
@@ -0,0 +1,512 @@
...
+    Returns:
+        str: The suitable subcountry name.
+    """
+    details = load_city_country_data_set()
+    for city_element in details:
+        if city_name.title() in city_element[CITY_NAME_KEY].title():
+            return city_element[SUBCOUNTRY_NAME_KEY_IN_CITIES]
+
+
+async def get_api_data(url: str) -> Optional[List[Any]]:
```
yammesicka, line 196: Cache this function to prevent high loading times

app/internal/world_clock.py
```diff
@@ -0,0 +1,512 @@
...
+        url (str):  The API url.
+    Returns:
+        list: The data.
+    """
+    try:
+        async with httpx.AsyncClient() as client:
+            resp = await client.get(url)
+            return resp.json()
+    except (json.JSONDecodeError, httpx.HTTPError) as errh:
+        print("Http Error:", errh)
```
yammesicka: Use logger, not print

app/internal/world_clock.py
```diff
@@ -0,0 +1,438 @@
...
+    """Get the subcountry of a city.
+    Args:
+        city_name (str): The given city name.
+    Returns:
+        str: The suitable subcountry name.
+    """
+    details = load_city_country_data_set()
+    for city_element in details:
+        if city_name.title() in city_element[CITY_NAME_KEY].title():
+            return city_element[SUBCOUNTRY_NAME_KEY_IN_CITIES]
```
yammesicka: Why not reformatting the JSON in a way that the city title would be the key?
Currently each search can take lots of time.

app/internal/world_clock.py
```diff
@@ -0,0 +1,438 @@
...
+    """Get the country of a city.
+    Args:
+        city_name (str): The given city name.
+    Returns:
+        str: The suitable country name.
+    """
+    details = load_city_country_data_set()
+    for city_element in details:
+        if city_name.title() in city_element[CITY_NAME_KEY].title():
+            return city_element[COUNTRY_NAME_KEY_IN_CITIES]
```
yammesicka: Same comment as get_subcountry

app/internal/world_clock.py
yammesicka: Use git precommit hooks to fix this formatting

yammesicka (changes requested):
Great job, please take a look at the insights :)

app/internal/world_clock.py
nir9696: I did my best to do so by using the decorator `@functools.lru_cache` but I always get runtime errors with the message: "cannot reuse already awaited coroutine". The truth is I'm pretty desperate and do not know how to solve this. I probably won't have the time to work on the code again by the end of the submission deadline - Can you consider giving up this matter or help me here? I have worked so hard...

app/internal/world_clock.py
nir9696: I don't see it. this note has already been resolved and I see the correct formmating without the `\` char.
This note is also "Outdated". Can you check again? :)

app/resources/country-continent.json
```diff
@@ -0,0 +1 @@
...State of", "Country_Number": 882, "Three_Letter_Country_Code": "WSM", "Two_Letter_Country_Code": "WS"},{"Continent_Code": "AS", "Continent_Name": "Asia", "Country_Name": "Yemen", "Country_Number": 887, "Three_Letter_Country_Code": "YEM", "Two_Letter_Country_Code": "YE"},{"Continent_Code": "AF", "Continent_Name": "Africa", "Country_Name": "Zambia, Republic of", "Country_Number": 894, "Three_Letter_Country_Code": "ZMB", "Two_Letter_Country_Code": "ZM"},{"Continent_Code": "OC", "Continent_Name": "Oceania", "Country_Name": "Disputed Territory", "Country_Number": null, "Three_Letter_Country_Code": null, "Two_Letter_Country_Code": "XX"},{"Continent_Code": "AS", "Continent_Name": "Asia", "Country_Name": "Iraq-Saudi Arabia Neutral Zone", "Country_Number": null, "Three_Letter_Country_Code": null, "Two_Letter_Country_Code": "XE"},{"Continent_Code": "AS", "Continent_Name": "Asia", "Country_Name": "United Nations Neutral Zone", "Country_Number": null, "Three_Letter_Country_Code": null, "Two_Letter_Country_Code": "XD"},{"Continent_Code": "AS", "Continent_Name": "Asia", "Country_Name": "Spratly Islands", "Country_Number": null, "Three_Letter_Country_Code": null, "Two_Letter_Country_Code": "XS"}]
```
Gonzom: Keys should be lowercase, reformat in a human readable way.
Also, is there really no python module that does this? I find this very very unlikely...

tests/test_world_clock.py
```diff
@@ -0,0 +1,263 @@
+from datetime import datetime
+import pytest
+
+import httpx
+import respx
+
+from app.internal.world_clock import (
```
yammesicka, line 7: Use git commit prehooks on this file

app/internal/world_clock.py
yammesicka: Have you tried [async-lru](https://github.com/aio-libs/async-lru)?

app/internal/world_clock.py
```diff
@@ -0,0 +1,510 @@
...
+
+def get_continent(country_name: str) -> Optional[str]:
+    """Get the continent of a given country.
+    Args:
+        country_name (str): The given country name.
+    Returns:
+        str: The suitable continent name.
+    """
+    details = load_country_continent_data_set()
+    for country_element in details:
```
yammesicka: Same here: put the name of the country as key to make the search much faster

app/internal/world_clock.py
```diff
@@ -0,0 +1,510 @@
...
+    Returns:
+        str: The normalized continent name.
+    """
+    for continent in CONTINENTS:
+        if continent in continent_name:
+            return continent
+    return continent_name
+
+
+def load_country_continent_data_set() -> List[Dict[str, str]]:
```
yammesicka, line 134: Cache the return value, reading and parsing JSON every time is expensive

app/internal/world_clock.py
```diff
@@ -0,0 +1,510 @@
...
+    details = load_country_continent_data_set()
+    for country_element in details:
+        if (country_name.title() in
+                country_element[COUNTRY_NAME_KEY_IN_CONTITNENTS]):
+            return normalize_continent_name(
+                country_element[CONTINENT_NAME_KEY]
+            )
+
+
+def load_city_country_data_set() -> List[Dict[str, str]]:
```
yammesicka, line 158: Cache the return value, reading and parsing JSON every time is expensive

app/internal/world_clock.py
```diff
@@ -0,0 +1,510 @@
...
+    Args:
+        place_name (str): The given place name.
+    Returns:
+        str: The suitable timezone.
+    """
+    res = [
+        timezone
+        for timezone in await get_api_data(TIMEZONES_BASE_URL)
+        if place_name in timezone
+    ]
```
yammesicka: This will cost an insane amount of time for no good reason :))
Move the `get_api_data` to a different line

app/internal/world_clock.py
```diff
@@ -0,0 +1,510 @@
...
+    Returns:
+        str: The timezone.
+    """
+    continent_url = f"{TIMEZONES_BASE_URL}/{place_name}"
+    api_data = await get_api_data(continent_url)
+    timezone = api_data[ARB_COUNTRY_TIMEZONE]
+    return f"{TIMEZONES_BASE_URL}/{timezone}"
+
+
+def load_country_subcountry_data_set() -> List[Dict[str, str]]:
```
yammesicka, line 376: Cache this

yammesicka (changes requested):

nir9696:
1. I tried to use both `async-lru`, for the asynchronous function, and ` functools.lru_cache`, for uploading the files,
and many tests fail suddenly for `RuntimeError: Event loop is closed`
and `174 Http Error` errors.
I tried everything and I am lost.
2. Also, my code is after the prehook fixes, so I do not know what I have left to fix ...

yammesicka:
There are few comments that aren't related to caching. Fix them and I'll consider this as a ticket :)

tests/test_world_clock.py
nir9696: Done :)

app/internal/world_clock.py
nir9696: done

app/internal/world_clock.py
nir9696: done

PR #234, opened 2021-02-07
Feature/unavailable user

noam-y:
creating a feature that allows disabling user. 
function will only allow disabling user if it doesnt own any event.
the front-end for disabling user will be added after the settings menu is merged :)


app/routers/user.py
```diff
@@ -59,3 +61,26 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
...
+
+
+def disable_user(session: Session, user_id: int):
+    """this functions changes user status to disabled.
+    returns:
+    True if function worked properly 
+    False if it didn't."""
+
+    events_user_owns = list(session.query(Event).filter_by(owner_id = user_id))
+    if len(events_user_owns) == 0:
```
yammesicka, line 73: Prefer a guard: `if events_user_owns: return False` and then the rest of the function, unindented.

app/routers/user.py
```diff
@@ -59,3 +61,26 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
...
+        """if user doesn't own any event, we will disable him and remove the user from all of its future events."""
+        user_disabled = session.query(User).get(user_id)
+        user_disabled.disabled = True
+        future_events_for_user = session.query(UserEvent.id).join(Event, Event.id == UserEvent.event_id).filter(UserEvent.user_id == user_id, Event.start > datetime.now()).all()
```
yammesicka, line 77: This line is too long, please split it

yammesicka (changes requested):
Good start :)

yammesicka:
Run `git rm -rf app/locales` and push again :)

tests/test_user.py
```diff
@@ -1,4 +1,52 @@
...
+
+    return user
+
+
+# -----------------------------------------------------
+# Tests
+# -----------------------------------------------------
+
+
+def test_disabling_user(new_user, session):
```
yammesicka, line 33: You should test all the possibilities :)

app/routers/user.py
```diff
@@ -59,3 +60,29 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
         session.query(Event).join(UserEvent)
         .filter(UserEvent.user_id == user_id).all()
     )
+
+
+def disable_user(session: Session, user_id: int):
```
yammesicka, line 65: Add return type

app/routers/user.py
```diff
@@ -59,3 +60,29 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
...
+
+    """if user doesn't own any event,
+    we will disable him and remove the user from all of its future events."""
+    user_disabled = session.query(User).get(user_id)
+    user_disabled.disabled = True
+    future_events_for_user = session.query(UserEvent.id).join(
+        Event, Event.id == UserEvent.event_id
+        ).filter(
+        UserEvent.user_id == user_id, Event.start > datetime.now()
+        ).all()
```
yammesicka, line 84: Hard to read

app/routers/user.py
```diff
@@ -59,3 +60,29 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
...
+    returns:
+    True if function worked properly
+    False if it didn't."""
+
+    events_user_owns = list(session.query(Event).filter_by(owner_id=user_id))
+    """if user owns any event, he won't be able to disable himself."""
+    if events_user_owns:
+        return False
+
+    """if user doesn't own any event,
```
yammesicka, line 76: Prefer # for short comments

app/routers/user.py
```diff
@@ -1,8 +1,9 @@
 from typing import List
 
+from datetime import datetime
```
yammesicka, line 3: Should be before `typing`, not in a new paragraph

app/routers/user.py
yammesicka, line 65: Can't we enable it?:)

yammesicka (changes requested):

app/routers/user.py
noam-y: the flake8 linter made me split lines way too many times. will documentation help?

app/routers/user.py
noam-y: oops! added enable_user func :)

tests/test_user.py
noam-y: Added seperate tests for the 3 different test cases :)

app/routers/user.py
```diff
@@ -59,3 +59,38 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
...
+    # we will disable him and remove the user from all of its future events.
+    user_disabled = session.query(User).get(user_id)
+    user_disabled.disabled = True
+    future_events_for_user = session.query(UserEvent.id).join(
+        Event, Event.id == UserEvent.event_id
+        ).filter(
+        UserEvent.user_id == user_id, Event.start > datetime.now()).all()
+    # the query above finds events that user participates in the future.
+    for event_connection in future_events_for_user:
+        session.delete(event_connection)
```
yammesicka, line 85: We shouldn't delete the user's events, he might want to get back in a minute

app/routers/user.py
```diff
@@ -59,3 +59,38 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
         session.query(Event).join(UserEvent)
         .filter(UserEvent.user_id == user_id).all()
     )
+
+
+def disable_user(session: Session, user_id: int) -> bool:
```
yammesicka, line 64: These functions look like an internal functions. Please move them to `app/internal` and add `/user/enable` and `/user/disable`.

Also, please check that only the user itself/the administrator does the disabling/enabling.

yammesicka (changes requested):

app/routers/user.py
noam-y: added this feature- but currently commented it since the logger system is not built yet, and I can't mock logged user without it :)

app/routers/user.py
noam-y: removed this part of the function. Added a func that returns all available users

app/internal/user/disable.py
```diff
@@ -0,0 +1,31 @@
+from datetime import datetime
```
yammesicka, line 1: Prefer to merge with `enable.py` to `availability.py` and call the functions `disable()` and `enable()` for the sake of simplicity :)

app/routers/user.py
```diff
@@ -2,7 +2,6 @@
```
yammesicka, line 2: What about a route to disable/enable the user? :)

yammesicka (changes requested):
Nice progress! :) Added few additional insights

app/routers/user.py
```diff
@@ -92,3 +96,55 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
...
+    if the disable fails user will stay at settings page
+    and an error will be shown."""
+    disable_successful = disable(session, get_current_user)
+    if disable_successful:
+        # disable succeeded- the user will be directed to homepage.
+        app.url_path_for("/")
+    else:
+        # if disable wasn't successful, user will stay in the
+        # settings page and an error message will show.
+        pass
```
yammesicka, line 133: Is this necessary?

app/routers/user.py
```diff
@@ -92,3 +96,55 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
...
+
+
+@router.post("/enable")
```
yammesicka, line 136: How is that different from the `/enable` above?

yammesicka (changes requested):

app/routers/user.py
noam-y: not really. deleted for now :)

app/routers/user.py
noam-y: above is a disabling request, this router is for enabling. used accidently the wrong function- fixed this in the current push

noam-y: oops. :/ accidently copied it instead of cutting for it to look nicer..

app/routers/user.py
```diff
@@ -1,13 +1,19 @@
...
+from starlette.status import HTTP_202_ACCEPTED
+
 from app.dependencies import get_db
 from app.database.models import Event, User, UserEvent
-from app.internal.utils import save
-from fastapi import APIRouter, Depends
+from app.internal.utils import save, get_current_user
+from app.internal.user.availability import (
+    disable, enable)
+
```
yammesicka, line 16: Sort by abc and split into 3 paragraphs, just as you did in app/internal/user/availability.py

yammesicka (changes requested):

app/routers/user.py
noam-y: fixed :D

noam-y: fixed :D

noam-y: fixed :D

app/routers/user.py
```diff
@@ -1,13 +1,17 @@
...
 
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session
 
 from app.dependencies import get_db
 from app.database.models import Event, User, UserEvent
-from app.internal.utils import save
-from fastapi import APIRouter, Depends
+from app.internal.user.availability import (
+    disable, enable)
```
yammesicka, line 12: Can be in 1 line

app/routers/user.py
```diff
@@ -92,3 +96,31 @@ def get_all_user_events(session: Session, user_id: int) -> List[Event]:
...
+        request: Request, session: Session = Depends(get_db)):
+    """route that sends request to disable the user.
+    after successful disable it will be directed to main page.
+    if the disable fails user will stay at settings page
+    and an error will be shown."""
+    disable_successful = disable(session, get_current_user)
+    if disable_successful:
+        # disable succeeded- the user will be directed to homepage.
+        url = router.url_path_for("home")
+        return RedirectResponse(url=url, status_code=HTTP_202_ACCEPTED)
```
yammesicka, line 112: Good job looking into HTTP codes, but [202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) is actually less preferable here.
Please change to 200 OK.

tests/test_user.py
```diff
@@ -1,4 +1,132 @@
...
+    """making sure only users who only participate in events
+    can disable and enable themselves."""
+    disable(session, user3.id)
+    assert user3.disabled
+    future_events = list(session.query(Event.id).join(UserEvent)
+                         .filter(
+                             UserEvent.user_id == user3.id,
+                             Event.start > datetime.now()
+                             ))
+    assert len(future_events) == 0
```
yammesicka, line 121: `assert not future_events`

tests/test_user.py
```diff
@@ -1,4 +1,132 @@
...
+def test_disabling_user_participating_event(session, user3):
+    """making sure only users who only participate in events
+    can disable and enable themselves."""
+    disable(session, user3.id)
+    assert user3.disabled
+    future_events = list(session.query(Event.id).join(UserEvent)
+                         .filter(
+                             UserEvent.user_id == user3.id,
+                             Event.start > datetime.now()
+                             ))
```
yammesicka, line 120: Reformat

tests/test_user.py
```diff
@@ -1,4 +1,132 @@
...
+
+def test_disabling_no_event_user(session, user1):
+    # users without any future event can disable themselves
+    disable(session, user1.id)
+    assert user1.disabled
+    future_events = list(session.query(Event.id).join(UserEvent)
+                         .filter(
+                             UserEvent.user_id == user1.id,
+                             Event.start > datetime.now()
+                             ))
```
yammesicka, line 104: Reformat

tests/test_user.py
```diff
@@ -1,4 +1,132 @@
...
+    assert not future_events
+    enable(session, user1.id)
+    assert not user1.disabled
+    # making sure that after disabling the user he can be easily enabled.
```
yammesicka, line 108: Should be above the relevant lines

yammesicka (changes requested):

tests/test_user.py
noam-y: not sure I reformatted this well :(

app/routers/user.py
```diff
@@ -1,13 +1,17 @@
...
 
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session
 
 from app.dependencies import get_db
 from app.database.models import Event, User, UserEvent
-from app.internal.utils import save
-from fastapi import APIRouter, Depends
+from app.internal.user.availability import disable, enable
+from app.internal.utils import save, get_current_user
```
yammesicka, line 13: `get_current_user, save`

app/internal/user/availability.py
```diff
@@ -0,0 +1,42 @@
...
+    # if get_current_user(session) != user_id:
+    #     return False
+    """line above makes sure the user disabled is the current user logged
+    & doesn't own any event.
+    currently it doesn't uses get_current_user since logger is not
+    merged yet, Ill add it when its impossible to mock a logged user."""
+
+    user_disabled = session.query(User).get(user_id)
+    user_disabled.disabled = True
+    session.merge(user_disabled)
```
yammesicka, line 28: Why here you do merge and in `enable` you don't? :)

app/internal/utils.py
```diff
@@ -43,6 +44,12 @@ def get_current_user(session: Session) -> User:
     return user
 
 
+def get_available_users(session: Session) -> List[User]:
+    """this function return all availible users."""
+    all_available_users = session.query(User).filter(not User.disabled).all()
+    return all_available_users
```
yammesicka, line 50: You can immediately return

app/routers/user.py
```diff
@@ -1,13 +1,17 @@
-from typing import List
+from fastapi import APIRouter, Depends, Request
 from pydantic import BaseModel, Field
+from starlette.responses import RedirectResponse
+from starlette.status import HTTP_200_OK
+from typing import List
```
yammesicka, line 5: typing should be in the 1st paragraph

app/routers/user.py
```diff
@@ -1,13 +1,17 @@
...
```
yammesicka, line 6: There shouldn't be a blank line here

yammesicka (changes requested):
Great job. Ready to merge, DM me with link when you fix the minor comments there :)

tests/test_user.py
```diff
@@ -1,4 +1,134 @@
...
+    )
+
+    association = UserEvent(
+        user_id=user.id,
+        event_id=event1.id
+    )
+    save(association, session)
+
+    return user
+
```
Gonzom, line 89: Can't the existing user and event fixtures work here instead of creating new ones?

tests/test_user.py
```diff
@@ -1,4 +1,134 @@
...
+
+# -----------------------------------------------------
+# Tests
+# -----------------------------------------------------
+
```
Gonzom, line 94: Please don't use this comment style.

app/internal/user/availability.py
```diff
@@ -0,0 +1,42 @@
+from datetime import datetime
+
```
Gonzom, line 2: Why is this file nested in a user directory? This should just be user.py

noam-y: yam asked for this :)

tests/test_user.py
noam-y: no since every user represents a different kind of test :(

app/internal/user/availability.py
noam-y: tried to do something to solve an error, it's solved already so deleted this

tests/test_user.py
Gonzom: Could you explain what is special about this user3 fixture that isn't found in the other User fixtures?

noam-y: user3 is a user that participates in an event that isn't owned by him. the seperation is important since if other tests add an event to this user, this test will fail.

tests/test_user.py
noam-y: why? 
i think it's very convenient; makes a nice seperation between fixtures and actual tests. seen this in other projects :)

Gonzom: It's not currently a style we are using here, nor is it supported by PEP 8. Not sure we should start a new style now for this.

tests/test_user.py
Gonzom: The user object doesn't have to be unique though. Fixtures should be reusable, if they aren't, then why are you creating fixtures?

User3, for example, is used only at test_disabling_user_participating_event().
In that test, just add the UserEvent code
```
   association = UserEvent(
        user_id=user.id,
        event_id=event1.id
    )
    save(association, session)
```
and then you have a user connected to events for your test.

Adding unnecessary fixtures pollutes the global fixtures and will cause bugs.

tests/test_user.py
noam-y: deleted! :)

tests/test_user.py
noam-y: understood your explanation, changed the code accordingly :)

PR #246, opened 2021-02-09
Fixed datetime dt error

annashtirberg:
Change with imports in last merge to develop make error:
undefined name "datetime"
when testing

I fixed it by correcting the use places where "datetime" was used to "dt"
very small fix and works :)


ivarshav (approved):

yammesicka:
Great initiative, thanks!

PR #253, opened 2021-02-10
Simplify agenda view

ivarshav:
Change duration view - 

Before:
image

After:
image


zohary89:
Hi!
Why do you prefer to simplify this view?
When I wrote this code, I prefered to present the duration in words instead of presenting the differences with unnecessary zeros...

ivarshav:
> Hi!
> Why do you prefer to simplify this view?
> When I wrote this code, I prefered to present the duration in words instead of presenting the differences with unnecessary zeros...

Great question! Thanks for asking it.
In my next issue, I'm going to filter in FE side the presented events by params.
Right now, it's much more complicated to do that since the objects in FE are complicated.

zohary89:
I understand, but I'm not sure why they are complicated.. It's basically a list of tuples, each one consist of the event and its duration in words (I used the "Arrow" module so the duration will be more readable).
Not sure it's best to change the functionality for simplification... can't we think of a better way to keep it and you'll still be able to build the FE filtering? :)

ivarshav:
> I understand, but I'm not sure why they are complicated.. It's basically a list of tuples, each one consist of the event and its duration in words (I used the "Arrow" module so the duration will be more readable).
> Not sure it's best to change the functionality for simplification... can't we think of a better way to keep it and you'll still be able to build the FE filtering? :)

This is more complicated since the structure is more complex. 
Also, I'm not sure it changes the functionality (only view).
I'm open to hear ideas and will be happy to discuss them :)

ivarshav:
> > I understand, but I'm not sure why they are complicated.. It's basically a list of tuples, each one consist of the event and its duration in words (I used the "Arrow" module so the duration will be more readable).
> > Not sure it's best to change the functionality for simplification... can't we think of a better way to keep it and you'll still be able to build the FE filtering? :)
> 
> This is more complicated since the structure is more complex.
> Also, I'm not sure it changes the functionality (only view).
> I'm open to hear ideas and will be happy to discuss them :)

@zohary89, I think I found a way to keep the same duration view & simplify the data structure :) will send a seperate PR.

zohary89:
Cool!  Thanks for the update me and good luck :)

ivarshav:
> Cool! Thanks for the update me and good luck :)

@zohary - https://github.com/PythonFreeCourse/calendar/pull/298 simplify structure and keeps duration format :)

Issue #255, opened 2021-02-11
Delete a double column


efratush:
Deleting a 'language' column in a user table and its use in a project.
Adding languages (English and Hebrew) to the database after creating the language table.

PR #256, opened 2021-02-11
Refactor/delete language column

efratush:
#255


efratush:
Mistake. I pushed to main instead of develop

PR #259, opened 2021-02-11
feat: statistics

hadaskedar2020:
This feature calculate statistics for given user and date-range.
It returns:
- Events count statistics - meetings_for_user & events that the user has created
- Day events statistics - minimum/maximum/average number of daily events the user has
- Events duration statistics in minutes - shortest/longest/average event the user has
- Participants statistics - maximum number of events the user has with same participant & relevant participant name


app/internal/statistics.py
```diff
@@ -0,0 +1,299 @@
...
+        start: start of date range.
+        end: end of date range.
+    Returns:
+        valid_input: boolean stating if input values are valid.
+        error_text: str (relevant ib case of error).
+        start: start of date range.
+        end: end of date range.
+    """
+    if not does_user_exist(session=db, user_id=userid):
+        return False, INVALID_USER, start, end
```
yammesicka: Prefer using NamedTuple for this

app/internal/statistics.py
```diff
@@ -0,0 +1,299 @@
...
+    if start is None:
+        start = datetime.datetime(year=datetime.datetime.now().year,
+                                  month=datetime.datetime.now().month,
+                                  day=datetime.datetime.now().day)
+    else:
+        start = datetime.datetime(year=start.year,
+                                  month=start.month,
+                                  day=start.day)
```
yammesicka:
```python
date = start or datetime.datetime.now()
start = datetime.datetime(date.year, date.month, date.day)
```

app/internal/statistics.py
```diff
@@ -0,0 +1,299 @@
...
+    if end is None:
+        end = start + datetime.timedelta(days=1, seconds=-1)
+    else:
+        end = datetime.datetime(year=end.year, month=end.month, day=end.day)\
+              + datetime.timedelta(days=1, seconds=-1)
```
yammesicka: Please don't split using `\`.

Another readable possibility: 

```python
single_day = datetime.timedelta(days=1, seconds=-1)
end = end or start
end_date = datetime.datetime(year=end.year, month=end.month, day=end.day)
day_after_end_date = end_date + single_day
```

app/internal/statistics.py
```diff
@@ -0,0 +1,299 @@
...
+    if start >= end:
+        return False, INVALID_DATE_RANGE, start, end
+    return True, "", start, end
+
+
+def get_events_count_stats(db: Session, userid: int, start: datetime.datetime,
+                           end: datetime.datetime)\
+        -> Dict[str, Dict[str, str]]:
```
yammesicka: Please don't split using `\`:

```python
def get_events_count_stats(
    db: Session, userid: int, start: datetime.datetime, end: datetime.datetime,
) -> Dict[str, Dict[str, str]]:
```

app/internal/statistics.py
```diff
@@ -0,0 +1,299 @@
...
+                   end: datetime.datetime)\
+        -> Tuple[bool, str, datetime.datetime, datetime.datetime]:
+    """ 1. input validations:
+            valid userid.
+            end date > start date.
+        2. date range preparation:
+            start: will be at 00:00 on input date (today if None).
+            end: will be at 23:59:59 on input date (today if None).
+    All statistics are based on full days (00:00:00-23:59:59).
+    Args:
```
yammesicka, line 58: Prefer putting one blank line before and after the `args` paragraph.

app/internal/statistics.py
yammesicka: `.count` returns string?

app/internal/statistics.py
```diff
@@ -0,0 +1,299 @@
...
+        db: db session.
+        userid: requested user id number for statistics.
+        start: start of date range.
+        end: end of date range.
+    Returns: json with
+        events_count_stats: (for requested date range).
+            meetings_for_user: events that the user has.
+            created_by_user: events that the user has created.
+    """
+    return {"events_count_stats":
```
yammesicka: Hard to read :(
Also, prefer a shallow structure if possible.

Export the queries to functions.
Also, WDYT about removing the outer "events_count_stats"?

yammesicka (changes requested):
That's awesome job!

Before I'll continue on checking the code, please take some time to look at the code formatting.
Avoid splitting lines using `\`, and do try to make it as readable as possible.

Usually, we want to allow people just skim through the code and understand it :)

app/internal/statistics.py
hadaskedar2020: cool!

app/internal/statistics.py
hadaskedar2020: Thanks.

app/internal/statistics.py
hadaskedar2020: Regarding export the queries to functions - did you mean to move it to the first part of the current function ar call a new one? (it the later case the current function doesn't do much) 
I gave it much thought and still believe that the outer "statistic area name" wrapping each few statistics makes the whole output much more orgenized and clear...
However, if you strongly believe that removing it is better let me know :-)

yammesicka: > Regarding export the queries to functions - did you mean to move it to the first part of the current function ar call a new one? (it the later case the current function doesn't do much)

I believe splitting it into two functions, or arranging the nesting a bit better, might help read this in just a quick skim-through.

> I gave it much thought and still believe that the outer "statistic area name" wrapping each few statistics makes the whole output much more orgenized and clear...

I must say I'm a great fan of "flat is better than nested". From my experience it greatly reduces the amount of bugs and simplifies the use of the system. I leave this choice to you though :)

hadaskedar2020: > I believe splitting it into two functions, or arranging the nesting a bit better, might help read this in just a quick skim-through.
> 
> >

I've rearranged the function - can you please take a look at the new code and advise me if any change is still required? (I'm not sure weather you ment that the "get_events_count_stats" function should call 1 or 2 other functions just for getting the data from the DB)

app/internal/statistics.py
```diff
@@ -0,0 +1,312 @@
...
+        start: start of date range.
+        end: end of date range.
+
+    Returns: NamedTuple with the following data:
+        valid_input: boolean stating if input values are valid.
+        error_text: str (relevant ib case of error).
+        start: start of date range.
+        end: end of date range.
+    """
+    Result = namedtuple(
```
yammesicka: Please use `typing.NamedTuple`, and put it in the global namespace

app/internal/statistics.py
```diff
@@ -0,0 +1,312 @@
...
+        'result', ['valid_input', 'error_text', 'start', 'end'])
+    if not does_user_exist(session=db, user_id=userid):
+        return Result(False, INVALID_USER, start, end)
```
yammesicka: `Result(valid_input=False, ...)`

app/internal/statistics.py
```diff
@@ -0,0 +1,312 @@
...
+    """
+    events_by_date = db.query(
+        func.date(Event.start), func.count(func.date(Event.start))).join(
+        UserEvent, UserEvent.event_id == Event.id).filter(
+        UserEvent.user_id == userid).filter(
+        or_(
+            and_(Event.start >= start, Event.start <= end),
+            and_(Event.end >= start, Event.end <= end),
+            and_(Event.start <= start, Event.end >= end))).group_by(
+        func.date(Event.start)).all()
```
yammesicka: Separate to a function

app/internal/statistics.py
```diff
@@ -0,0 +1,312 @@
...
+    if len(events_by_date) > 0:
+        min_events_in_day = min(day[1] for day in events_by_date)
+        max_events_in_day = max(day[1] for day in events_by_date)
+        sum_events_per_period = sum(day[1] for day in events_by_date)
+    else:
+        min_events_in_day = 0
+        max_events_in_day = 0
+        sum_events_per_period = 0
+    if num_of_days_in_period > len(events_by_date):
+        min_events_in_day = 0
```
yammesicka: Separate to a function, return as namedtuple

app/internal/statistics.py
```diff
@@ -0,0 +1,312 @@
...
+        (func.max(func.julianday(Event.end) - func.julianday(Event.start))
+         * NIN_IN_DAY),
+        (func.avg(func.julianday(Event.end) - func.julianday(Event.start))
+         * NIN_IN_DAY)
+    ).join(UserEvent, UserEvent.event_id == Event.id).filter(
+        UserEvent.user_id == userid).filter(
+        or_(
+            and_(Event.start >= start, Event.start <= end),
+            and_(Event.end >= start, Event.end <= end),
+            and_(Event.start <= start, Event.end >= end))).all()
```
yammesicka: Separate to a function.
Also, the last 4 lines returns many time in your code. I'd export them to a function as well.

yammesicka (changes requested):
Keep up the good work :)

app/internal/statistics.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+ValidateResult = namedtuple(
+    'ValidateResult', 'valid_input error_text start end')
+DailyEventsStatistics = namedtuple(
+    'DailyEventsStatistics',
+    'min_events_in_day max_events_in_day avg_events_per_day')
+EventsDurationStatistics = namedtuple(
+    'EventsDurationStatistics',
+    'shortest_event longest_event average_event')
```
yammesicka: Please use typing.NamedTuple

app/internal/statistics.py
```diff
@@ -0,0 +1,391 @@
...
+        end: end of date range.
+
+    Returns:
+        data of date + number of events on it
+    """
+    events_by_date = db.query(
+        func.date(Event.start), func.count(func.date(Event.start))).join(
+        UserEvent, UserEvent.event_id == Event.id).filter(
+        UserEvent.user_id == userid).filter(
+        text(event_start_end_filter())).all()
```
yammesicka: We always prefer to split the code in a fashion that will allow others to easily skim through it.

Compare with:
```python
def get_date_filter_between_dates(start: datetime.datetime, end: datetime.datetime) -> ?:
    """
    Prepare the filter by dates using declarative SQLAlchemy.

    Returns:
        concatenated sql part for events date
    """
    event_start, event_end = func.date(Event.start), func.date(Event.end)
    return (
        (Event.start >= start and Event.start <= end)
        or (Event.end >= start and Event.end <= end)
        or (Event.start <= start and Event.end >= end)
    )

def get_events_by_date(
        db: Session, userid: int, start: datetime.datetime,
        end: datetime.datetime) -> List[Tuple[datetime.datetime, int]]:
    """Get (what?) about the date, and number of events on it.

    Args:
        db: db session.
        userid: requested user id number for statistics.
        start: start of date range.
        end: end of date range.

    Returns:
        Number of events on the date range.
    """
    start_date = func.date(Event.start)
    events_count = func.count(start_date)
    by_user_id = UserEvent.user_id == userid
    user_to_event = (UserEvent, UserEvent.event_id == Event.id)
    return (
        db.query(start_date, events_count).join(user_to_event)
        .filter(by_user_id)
        .filter(get_date_filter_between_dates(start, end))
        .all()
    )

File: app/internal/statistics.py [hunk 14]
```diff
@@ -0,0 +1,391 @@
...
+    Returns:
+        NamedTuple of: shortest_event, longest_event, average_event
+    """
+    events_duration_statistics = db.query(
+        (func.min(func.julianday(Event.end) - func.julianday(Event.start))
+         * NIN_IN_DAY),
+        (func.max(func.julianday(Event.end) - func.julianday(Event.start))
+         * NIN_IN_DAY),
+        (func.avg(func.julianday(Event.end) - func.julianday(Event.start))
+         * NIN_IN_DAY)
```
yammesicka: Save `func.julianday(Event.end) - func.julianday(Event.start)` on a variable outside the `db.query`.  It will help this lines be much more readable

yammesicka (changes requested):
Another round of reformating :)

app/internal/statistics.py
```diff
@@ -0,0 +1,395 @@
...
+        ('min_events_in_day', int), ('max_events_in_day', int),
+        ('avg_events_per_day', float)])
+EventsDurationStatistics = NamedTuple('EventsDurationStatistics', [
+        ('shortest_event', float), ('longest_event', float),
+        ('average_event', float)])
+
+
+def validate_input(
+        db: Session, userid: int, start: datetime.datetime,
+        end: datetime.datetime) -> NamedTuple('ValidateResult'):
```
yammesicka: Just `ValidateResult`

app/internal/statistics.py
```diff
@@ -0,0 +1,395 @@
...
+from app.database.models import Event, UserEvent
+from app.routers.user import does_user_exist, get_users
+
+SUCCESS_STATUS = 0
+ERROR_STATUS = -1
+INVALID_DATE_RANGE = "End date must be later than start date"
+INVALID_USER = "Invalid user id"
+NIN_IN_DAY = 1440
+
+ValidateResult = NamedTuple('ValidateResult', [
```
yammesicka: Prefer: `ValidationResult` (class names should be adjectives and not verbs)

app/internal/statistics.py
```diff
@@ -0,0 +1,395 @@
...
+    max_events_in_day = max(day[1] for day in events_by_date)
+    sum_events_per_period = sum(day[1] for day in events_by_date)
+    if num_of_days_in_period > len(events_by_date):
+        min_events_in_day = 0
+    avg_events_per_day = round(
+        sum_events_per_period / num_of_days_in_period, 2)
+    return DailyEventsStatistics(
+        min_events_in_day=min_events_in_day,
+        max_events_in_day=max_events_in_day,
+        avg_events_per_day=avg_events_per_day)
```
yammesicka: You can leave out the keyword arguments in case of x=x

yammesicka (changes requested):
Great job :) Let's give it the last push

tests/test_statistics.py
```diff
@@ -0,0 +1,73 @@
+import datetime
+
+from app.internal.statistics import get_statistics
+from app.internal.statistics import INVALID_DATE_RANGE, INVALID_USER,\
+    SUCCESS_STATUS
```
yammesicka: Don't split using \

yammesicka (changes requested):
.

app/internal/statistics.py
```diff
@@ -0,0 +1,395 @@
...
+            stat_3: display summary for "events_duration_statistics".
+            stat_4: display summary for "participants_statistics".
+    """
+    output = {}
+    validate_result = validate_input(db, userid, start, end)
+    if not validate_result.valid_input:
+        output["start"] = validate_result.start
+        output["end"] = validate_result.end
+        output["Status"] = ERROR_STATUS
+        output["ErrorDescription"] = validate_result.error_text
```
Gonzom: User lowercase names - "status" and "Error_description"

app/internal/statistics.py
```diff
@@ -0,0 +1,395 @@
...
+from typing import Dict, List, NamedTuple, Tuple, Union
+
+from sqlalchemy import and_, func, or_
+from sqlalchemy.orm import Session
+from sqlalchemy.util import symbol
+
+from app.database.models import Event, UserEvent
+from app.routers.user import does_user_exist, get_users
+
+SUCCESS_STATUS = 0
```
Gonzom: Can this be something other than 0 or 1? If it can't, make it a boolean.

app/internal/statistics.py
yammesicka: ? Not really solved

app/internal/statistics.py
yammesicka: Without the NamedTuple

app/internal/statistics.py
yammesicka: ?

yammesicka (changes requested):
There are many issues that were not resolved. Please revise again

hadaskedar2020:
> There are many issues that were not resolved. Please revise again
For a strange reason (i'm trying to figure out now) all my last changes were not commited to Github :-(
I'm very sorry for this problem and for your lost time in reviewing this.
I will figure out the problem and commit it again.

app/internal/statistics.py
```diff
@@ -0,0 +1,393 @@
...
+    if not validate_result.valid_input:
+        output["start"] = validate_result.start
+        output["end"] = validate_result.end
+        output["status"] = ERROR_STATUS
+        output["error_description"] = validate_result.error_text
+        return output
+    output["status"] = SUCCESS_STATUS
+    output["error_description"] = ""
+    output.update(get_events_count_stats(
+        db, userid, validate_result.start, validate_result.end))
```
yammesicka: Put parameters into a list and pass them using `*args`

app/internal/statistics.py
```diff
@@ -0,0 +1,393 @@
...
+    validate_result = validate_input(db, userid, start, end)
+    if not validate_result.valid_input:
+        output["start"] = validate_result.start
+        output["end"] = validate_result.end
+        output["status"] = ERROR_STATUS
+        output["error_description"] = validate_result.error_text
+        return output
+    output["status"] = SUCCESS_STATUS
+    output["error_description"] = ""
+    output.update(get_events_count_stats(
```
yammesicka: Put functions into list and loop over them

app/internal/statistics.py
```diff
@@ -0,0 +1,393 @@
...
+        db, userid, validate_result.start, validate_result.end))
+    output.update(get_daily_events_statistics(
+        db, userid, validate_result.start, validate_result.end))
+    output.update(get_events_duration_statistics(
+        db, userid, validate_result.start, validate_result.end))
+    output.update(get_participants_statistics(
+        db, userid, validate_result.start, validate_result.end))
+    output.update(prepare_display_text(
+        output, validate_result.start, validate_result.end))
```
yammesicka: Prefer separating this part into another functions

app/internal/statistics.py
```diff
@@ -0,0 +1,393 @@
...
+    user_to_event = (UserEvent, UserEvent.event_id == Event.id)
+    participant_count = func.count(UserEvent.user_id)
+    subquery = db.query(
+        Event.id).join(user_to_event).filter(by_user_id).filter(
+        get_date_filter_between_dates(start, end)).subquery()
+    event_participants = db.query(
+        UserEvent.user_id, participant_count).filter(
+        by_not_user_id).filter(
+        UserEvent.event_id.in_(subquery)).group_by(
+        UserEvent.user_id).order_by(participant_count.desc()).first()
```
yammesicka: Use pre-commit hooks to automatically fix the formatting here (there is an explanation in the README)

yammesicka (commented):
Great progress :) Please review the responses to your last commit

app/internal/statistics.py
```diff
@@ -0,0 +1,481 @@
...
+    ],
+)
+EventsDurationStatistics = NamedTuple(
+    "EventsDurationStatistics",
+    [
+        ("shortest_event", float),
+        ("longest_event", float),
+        ("average_event", float),
+    ],
+)
```
yammesicka, line 41: Perfect! The indentation are just beautiful now. Much more readable :)

app/internal/statistics.py
```diff
@@ -0,0 +1,481 @@
...
+        2. date range preparation:
+            start: will be at 00:00 on input date (today if None).
+            end: will be at 23:59:59 on input date (today if None).
+    All statistics are based on full days (00:00:00-23:59:59).
+
+    Args: list of parameters:
+        params[0] - db: db session.
+        params[1] - userid: requested user id number for statistics.
+        params[2] - start: start of date range.
+        params[3] - end: end of date range.
```
yammesicka: So what not having `db, user_id, start, end` in the parameter list?

app/internal/statistics.py
```diff
@@ -0,0 +1,481 @@
...
+    by_not_user_id = UserEvent.user_id != params[1]
+    user_to_event = (UserEvent, UserEvent.event_id == Event.id)
+    participant_count = func.count(UserEvent.user_id)
+    subquery = (
+        params[0]
+        .query(Event.id)
+        .join(user_to_event)
+        .filter(by_user_id)
+        .filter(get_date_filter_between_dates(params[2], params[3]))
+        .subquery()
```
yammesicka, line 365: Is this needed?

yammesicka (commented):
Almost ready :)

app/internal/statistics.py
hadaskedar2020: I think that since we use a query on a result of another query:
The sub query gets all relevant events for the user (owner or participant), while the outer query gets participants for those events.
Therefore, in  my opinion, this syntax suits this need.

PR #262, opened 2021-02-11
Feature/separate vc link

Michael-B-D:
separate the VC link from the location in the frontend and in the DB


Michael-B-D:
open new pr

PR #266, opened 2021-02-12
Few changes- agenda html

zohary89:
Fix link to the event view
Add event color as a background
few changes in css


app/templates/agenda.html
```diff
@@ -6,51 +6,58 @@
...
-
-  <div class="container mt-4">
+  <div class="pt-4 px-5">
     {% if start_date > end_date %}
       <p>{{ gettext("Start date is greater than end date") }}</p>
     {% elif events | length == 0 %}
       <p>{{ gettext("No events found...") }}</p>
     {% elif start_date == end_date %}
-      <h1>{{ start_date.strftime("%d/%m/%Y") }}</h1>
+      <h1 id="dates">{{ start_date.strftime("%d/%m/%Y") }}</h1>
```
yammesicka, line 41: If we already at it, can you please change it to h2 and have another element with h1? :)

There should be only 1 h1 in page

yammesicka (changes requested):

app/templates/agenda.html
zohary89: The page will eventually contain only 1 h1 element, as the two h1 elements are part of a condition.
Is it ok or should I change it?
Maybe it's better to use regular <span>, because I don't have another element that can be h1?

PR #271, opened 2021-02-12
feature/event privacy + fix eventview error

imimouni:
3 levels of event privacy-
'Public' - can be shown by anyone
'Private' - for other users - The event start date end date and owner is shown all other details are 'private' or not shown
'Hidden' - can only be seen by owner, the event will not be shown to anyone other than the owner

small fix in eventview function - a '?' was missing between 'event_id' and 'messages=' in the routing link.


app/routers/event.py
```diff
@@ -16,11 +17,17 @@
...
+# TODO add this when the user system is merged (PR#195)
+# from app.internal.security.dependancies import (
+# current_user, CurrentUser)
 from app.internal.utils import create_model
 from app.routers.user import create_user
 
+
 TIME_FORMAT = '%Y-%m-%d %H:%M'
 
+PRIVATE = 'Private'
```
yammesicka: Maybe we should have an `enum` for `PrivacyKinds`?

app/routers/event.py
```diff
@@ -72,32 +82,100 @@ async def create_new_event(request: Request, session=Depends(get_db)):
...
+        status_code=status.HTTP_404_NOT_FOUND,
+        detail=error_message)
+
+
 @router.get("/{event_id}")
 async def eventview(request: Request, event_id: int,
                     db: Session = Depends(get_db)):
     event = by_id(db, event_id)
+    event_to_show = can_show_event(event, db)
+    if not event_to_show:
```
yammesicka: Small suggestion which is not that important: `if not can_show_event(event, db)` looks great :)

app/routers/event.py
```diff
@@ -72,32 +82,100 @@ async def create_new_event(request: Request, session=Depends(get_db)):
...
     return templates.TemplateResponse("event/eventview.html",
-                                      {"request": request, "event": event,
+                                      {"request": request,
+                                       "event": event_to_show,
                                        "start_format": start_format,
                                        "end_format": end_format,
                                        "messages": messages})
 
 
+class PrivateEvent:
```
yammesicka: Maybe we should put this in `internal/privacy`?

app/routers/event.py
```diff
@@ -112,11 +190,7 @@ def by_id(db: Session, event_id: int) -> Event:
     try:
         event = db.query(Event).filter_by(id=event_id).one()
     except NoResultFound:
-        error_message = f"Event ID does not exist. ID: {event_id}"
-        logger.exception(error_message)
-        raise HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
-            detail=error_message)
+        nonexisting_event(event_id)
```
yammesicka: Please leave the logging in tact :)

yammesicka (commented):
Great job! I've added few insights, please take a look :)

app/routers/event.py
imimouni: I didn't touch it, just moved it to a separate function so that I can reuse it - is that not ok?

yammesicka: Ah, I haven't noticed. kk

app/routers/event.py
imimouni: I actually used "event_to_show" later in the code again (line 117 - in the template response) so it was useful to do this like that :)

app/routers/event.py
imimouni: Added in internal/privacy.py and updated in all relevant places

app/routers/event.py
imimouni: Done :)

app/routers/event.py
```diff
@@ -257,38 +316,45 @@ def update_event(event_id: int, event: Dict, db: Session
...
+                         content=content,
+                         owner_id=owner_id,
+                         location=location,
+                         vc_link=vc_link,
+                         color=color,
+                         emotion=get_emotion(title, content),
+                         invitees=invitees_concatenated,
+                         category_id=category_id,
+                         availability=availability,
+                         is_google_event=is_google_event)
```
yammesicka: I actually prefer it the previous way

app/routers/event.py
```diff
@@ -70,65 +77,129 @@ async def create_event_api(event: EventModel, session=Depends(get_db)):
...
+                         title=title,
+                         start=start,
+                         end=end,
+                         owner_id=owner_id,
+                         privacy=privacy,
+                         content=content,
+                         location=location,
                          invitees=invited_emails,
                          category_id=category_id,
                          availability=availability)
```
yammesicka:
```
create_events(
    session, title, start, end, owner_id, privacy, content, location,
    invited_emails, category_id, availability,
)
```

(less space, without repeating things, easy enough to see git diif on next addition)

app/internal/privacy.py
```diff
@@ -0,0 +1,25 @@
...
+        self.start = start
+        self.end = end
+        self.privacy = PrivacyKinds.Private.name
+        self.content = PrivacyKinds.Private.name
+        self.owner_id = owner_id
+        self.location = PrivacyKinds.Private.name
+        self.color = Null
+        self.invitees = PrivacyKinds.Private.name
+        self.category_id = Null
+        self.emotion = Null
```
yammesicka: Why aren't we using the regular `Event` object from the ORM?

app/routers/event.py
```diff
@@ -70,65 +77,129 @@ async def create_event_api(event: EventModel, session=Depends(get_db)):
...
+        private_event = PrivateEvent(
+            start=event.start,
+            end=event.end,
+            owner_id=event.owner_id,
+        )
+        return private_event
+    elif event.privacy == PrivacyKinds.Hidden.name and not is_owner:
+        return
+    elif event.privacy == PrivacyKinds.Public.name or is_owner:
+        return event
```
yammesicka: According to the function name, I'd expect to see here True/False reponse.

app/routers/event.py
```diff
@@ -303,20 +369,19 @@ def sort_by_date(events: List[Event]) -> List[Event]:
...
         by event id."""
-    return [email[0] for email in
-            db.query(User.email).select_from(Event).join(
-                UserEvent, UserEvent.event_id == Event.id).join(
+    return [
+        email[0] for email in db.query(User.email).select_from(Event).join(
+            UserEvent, UserEvent.event_id == Event.id).join(
                 User, User.id == UserEvent.user_id).filter(
-                Event.id == event_id).all()]
+                    Event.id == event_id).all()
```
yammesicka: Unreadable. Can you please reformat?

app/routers/event.py
```diff
@@ -347,8 +410,7 @@ def delete_event(event_id: int,
         pass
         # TODO: Send them a cancellation notice
         # if the deletion is successful
-    return RedirectResponse(
-        url="/calendar", status_code=status.HTTP_200_OK)
+    return RedirectResponse(url='/calendar', status_code=status.HTTP_200_OK)
```
yammesicka: Prefer url_for

tests/test_event.py
```diff
@@ -4,17 +4,18 @@
...
 from sqlalchemy.orm.session import Session
 from starlette import status
 
 from app.database.models import Comment, Event
 from app.dependencies import get_db
+from app.internal.privacy import PrivacyKinds
 from app.internal.utils import delete_instance
 from app.main import app
-
+from app.routers.event import can_show_event
```
yammesicka: Should be after `app.routers`

yammesicka (changes requested):
Great job! I've added few more insights

app/routers/event.py
imimouni: I don't know why this happened...
Almost for all of your comments are stuff I didn't actually change . It's from the prehooks that automatically formatted everything 😭

app/internal/privacy.py
imimouni: A few reasons - as to not modify the original event to hide all the details, to not create an extra redundant instance in the database and also when using the Event instance there is a problem with an sa_instance attribute that created a lot of problems when I tried to create a copy of the original event.
In short I think that using this neutral object will prevent messing up and cluttering the database and provide an event element with only the relevant non private details.

app/routers/event.py
imimouni: Ill try... i didnt actually change it it was theprehooks autoformat 😭
Its not even relevant to my feature

app/routers/event.py
imimouni: Not related to my feature this is from the merge.

app/routers/event.py
yammesicka: Good to know! I've updated the prehook for you to use a better autoformatter. Try now.

app/internal/privacy.py
yammesicka: There's a major problem with such an object, 'cause we will need to keep it updated for any changes of the original event object.
Don't you create it from existing event anyway?

imimouni: Do you have a better suggestion ?
Originally I tried to create a copy of the original event but this resulted in a lot of errors as sqlalchemy adds an "sa_instance" attribute to the object, and when I tried to remove it only for the event copy that also cause a lot of errors. I researched this a lot and found that this is the way others solved these issues- by creating another object to represent only the values we need without a connection to the database.

Is there some way that I could make sure this object will accept any future attributes that will be added to the Event table? Like maybe make it inherit from it ?

app/routers/event.py
imimouni: Changed it back to the way it was.

app/routers/event.py
imimouni: Changed it back (as mentioned the prehooks and not me :/ )

app/routers/event.py
imimouni: changed function name to "event_to_show" which makes more sense as this returns an event (or nothing if the event is not allowed to be shown).
changed the "event_to_show" variable name accordingly.

tests/test_event.py
imimouni: Thanks! fixed

app/routers/event.py
imimouni: Thank you, in any case I changed it back to the original way it was before the auto-formatting

app/routers/event.py
imimouni: So, I know you said that it's preferable to not repeat things but as the function itself has some deafult values that sometimes are necessary or not I changed the formatting but did leave the explicit attributing of each variable to the parameters as I saw that otherwise this causes problems and misunderstanding in tests and so on.
It's less pretty yes, but leaves less room for mistakes and doesn't require to change the create_event function parameters order (especially parameters that are not related to my feature which could confuse or mess up other things)

app/routers/event.py
imimouni: BTW the most recent commit includes the new prehook so please notice that some changes result from that and not from me :)`

app/internal/privacy.py
yammesicka: > Originally I tried to create a copy of the original event but this resulted in a lot of errors as sqlalchemy adds an "sa_instance" attribute to the object

Why is that a problem?

---

I think the best solution is to use the original `Event` instance.
If it is not possible, maybe we should inherit from it.

imimouni: ok I found out (lots of trial and error and additional research) how to overcome sqlalchemy weird additional attributes and create a copy of the event as an Event object, without saving it :)
will be in the commit I will soon upload

imimouni: Now uploaded :)

app/routers/event.py
```diff
@@ -104,8 +108,12 @@ async def create_new_event(
 
     vc_link = data["vc_link"]
     category_id = data.get("category_id")
+    privacy = data["privacy"]
+    privacy_kinds = [kind.name for kind in PrivacyKinds]
+    if privacy not in privacy_kinds:
+        privacy = PrivacyKinds.Public.name
     is_google_event = data.get("is_google_event", "True") == "True"
-
+    print(is_google_event)
```
yammesicka: Remove the print

app/templates/event/partials/edit_event_details_tab.html
```diff
@@ -1,86 +0,0 @@
-<div class="form_row">
```
yammesicka: Did you removed this on purpose?

yammesicka (changes requested):
Great job. We're about to finish working on this feature. DM me when you finish

app/templates/event/partials/edit_event_details_tab.html
imimouni: Well yes.
I noticed the file was moved to templates/partials/calendar/event so I thought this was left over accidentally as it wasn't being loaded from anywhere.

app/routers/event.py
imimouni: thanks! sorry :) fixed

PR #292, opened 2021-02-15
Feature/meds reminder

ShiZinDle:
Add automatic medication reminder feature (code, html, tests).
Moved `get_time_from_string` from app.salary.utils to app.internal.utils
Fixed jinja2 templating base (removed duplicate 'base.html' and switched all extends to 'partials/index/index_base.html')


app/routers/event.py
```diff
@@ -79,7 +79,7 @@ async def create_new_event(request: Request,
 
     messages = get_messages(session, event, uninvited_contacts)
     return RedirectResponse(router.url_path_for('eventview', event_id=event.id)
-                            + f'messages={"---".join(messages)}',
+                            + f'?messages={"---".join(messages)}',
```
PureDreamer: Why the ? sign?

app/routers/meds.py
```diff
@@ -0,0 +1,364 @@
...
+    end_date = get_time_from_string(form['end'])
+    amount = int(form['amount'])
+    early = get_time_from_string(form['early'])
+    late = get_time_from_string(form['late'])
+    minimum = get_time_from_string(form['min'])
+    maximum = get_time_from_string(form['max'])
+    first_time = first if first else early
+    start = datetime.combine(start_date, first_time)
+    if late <= early:
+        end_date += timedelta(1)
```
PureDreamer: shouldnt it be timedelta(days=1) ?

app/templates/partials/index/navigation.html
```diff
@@ -30,6 +30,15 @@
...
           <a class="nav-link" href="/search">Search</a>
         </li>
+        <li class="nav-item">
+          <a class="nav-link" href="/celebrity">Celebrities Born Today</a>
+        </li>
+        <li class="nav-item">
+          <a class="nav-link" href="{{ url_for('salary_home') }}">Salary</a>
+        </li>
+        <li class="nav-item">
+          <a class="nav-link" href="{{ url_for('meds') }}">Medication</a>
```
PureDreamer: I think it is more suited for the feature panal by @aviadamar what do you think?

PureDreamer (commented):
Very good PR, hope it will be merged soon!

app/routers/event.py
ShiZinDle: This is the way to pass additional query parameters in `get` requests ([Take a look at the syntax portion of this wikipedia article about URLs](https://en.wikipedia.org/wiki/URL)). The redirection didn't work without this addition.

app/routers/meds.py
ShiZinDle: This would be a more "polite" and clear way to write it, but since `days` is the first parameter of `datetime.timedelta` it is not required to explicitly specify it.

app/templates/partials/index/navigation.html
ShiZinDle: I reckon all of the links in this section should eventually transfer to the feature panel, but since it is currently only visible on one page, I put the link here with the rest of the features.

app/routers/event.py
PureDreamer: ok, thanks for the info!

app/internal/utils.py
```diff
@@ -57,3 +59,25 @@ def get_user(session: Session, user_id: int) -> Optional[User]:
...
+    try:
+        return datetime.strptime(string, '%Y-%m-%d').date()
+    except ValueError:
+        try:
+            return datetime.strptime(string, '%H:%M').time()
+        except ValueError:
+            try:
+                return datetime.strptime(string, '%H:%M:%S').time()
+            except ValueError:
+                return None
```
yammesicka: WDYT about:

```python
formats = {'%Y-%m-%d': 'date', '%H:%M': 'time', %H:%M:%S': 'time'}
for time_format, method in formats.items():
    try:
        time_obj = getattr(datetime.strptime(string, time_format), method)
    except ValueError:
        pass
    else:
        return time_obj
```

app/routers/meds.py
yammesicka: You are both right. I personally would love to see the `days` argument here, it'll make it much easier to understand.

app/routers/meds.py
```diff
@@ -0,0 +1,364 @@
...
+    Args:
+        early (time): Earlier time object.
+        late (time): Later time object. Interpreted as following day if earlier
+        than early.
+
+    Returns:
+        int: Interval between time objects in minutes.
+    """
+    if early == late:
+        0
```
yammesicka: Oops?

app/routers/meds.py
```diff
@@ -0,0 +1,364 @@
...
+    extra = 1 if early > late else 0
```
yammesicka: Can also write: `int(early > late)`

app/routers/meds.py
```diff
@@ -0,0 +1,364 @@
...
+
+    Returns:
+        bool: True if interval is sufficient for reminder amount with minimal
+              interval constraint, False otherwise
+    """
+    min_time = (amount - 1) * convert_time_to_minutes(minimum)
+    interval = get_interval_in_minutes(early, late)
+    if min_time <= interval:
+        return True
+    return False
```
yammesicka: return min_time <= interval

app/routers/meds.py
```diff
@@ -0,0 +1,364 @@
...
+def calc_reminder_interval_in_seconds(form: Dict[str, str]) -> int:
+    """Returns interval between reminders in seconds.
+
+    Args:
+        form (dict(str, str)): Medication form containing all relevant data.
+
+    Returns:
+        int: Interval between reminders in seconds.
+    """
+    (_, _, amount, early, late, minimum, maximum, _, _, _) = trans_form(form)
```
yammesicka: No need to include parenthesis here

app/routers/meds.py
```diff
@@ -0,0 +1,364 @@
...
+              a given date or equal to any of them, False otherwise.
+    """
+    early_datetime = datetime.combine(day, early)
+    late_datetime = datetime.combine(day, late)
+    if late < early:
+        late_datetime += timedelta(1)
+    return early_datetime <= reminder <= late_datetime
+
+
+def get_first_day_reminders(form: Dict[str, str],
```
yammesicka: Can you please split this function? :) It's a little bit heavy

app/routers/meds.py
```diff
@@ -0,0 +1,364 @@
...
+                    reminder = datetimes[-1] + timedelta(
+                        minutes=min_minutes)
+                    if validate_datetime(reminder,
+                                         datetimes[-1].date(),
+                                         early, late):
+                        datetimes.append(reminder)
+    return datetimes
+
+
+def get_reminder_datetimes(form: Dict[str, str]) -> List[datetime]:
```
yammesicka: Split, if possible

yammesicka (changes requested):
Great start! Added few insights :)

app/internal/utils.py
ShiZinDle: I tried to avoid passing on an exception, but I can't say I'm happy with how messy it became. I'll implement your version.

app/routers/meds.py
ShiZinDle: Great trick. Thx.

app/routers/meds.py
ShiZinDle: Sometimes we try to make the most simple things appear more difficult than they actually are.
Sometimes we're just missing silly mistakes :D

app/routers/meds.py
```diff
@@ -0,0 +1,424 @@
...
+    late = get_time_from_string(form['late'])
+    minimum = get_time_from_string(form['min'])
+    maximum = get_time_from_string(form['max'])
+    first_time = first if first else early
+    start = datetime.combine(start_date, first_time)
+    end_date = adjust_day(end_date, early, late, eq=True)
+    end = datetime.combine(end_date, late)
+    note = form['note']
+    return (
+        name, first, amount, early, late, minimum, maximum, note, start, end)
```
yammesicka: For future development support, it probably would be more convenient to get those values back in a dictionary.
That way the programmer who uses `trans_form` would be able to use `form.get('end_date')`, instead of having to go to this function and unpack all the values.

app/routers/meds.py
```diff
@@ -0,0 +1,424 @@
...
+    if amount == 1:
+        return 0
+    reminder_interval = get_interval_in_minutes(early, late)
+    max_med_interval = reminder_interval / (amount - 1)
+    min_minutes = convert_time_to_minutes(minimum)
+    max_minutes = convert_time_to_minutes(maximum)
+    avg_med_interval = (min_minutes + max_minutes) / 2
+    if avg_med_interval >= max_med_interval:
+        return int(max_med_interval * 60)
+    return int(avg_med_interval * 60)
```
yammesicka: `return int(min(avg_med_interval, max_med_interval) * 60)`

app/routers/meds.py
```diff
@@ -0,0 +1,424 @@
...
+            'start': event_time,
+            'end': event_time + timedelta(minutes=5),
+            'content': note,
+            'owner_id': user_id,
+        }
+        events.append(create_model(session, Event, **data))
+
+
+@router.get('/')
+@router.post('/')
```
yammesicka, line 24: I think this is the only function which should stay in `routers/`, the rest should go into `/internals`

app/routers/meds.py
```diff
@@ -0,0 +1,424 @@
...
+async def meds(request: Request,
+               session: Session = Depends(get_db)) -> Response:
+    """Renders medication reminders creation form page. Creates reminders in DB
+    and redirects to home page upon submition if valid."""
+    form = await request.form()
+    errors = []
+
+    data = {
```
yammesicka: `data` is often considered a bad variable name.
Can you please give it a better name?

app/routers/meds.py
```diff
@@ -0,0 +1,424 @@
...
+        'max': time(23, 59),
+        'note': '',
+    }
+
+    if form:
+        user = get_current_user(session)
+        errors = validate_form(form)
+        if not errors:
+            create_events(session, user.id, form)
+            return RedirectResponse(app.url_path_for('home'), status_code=303)
```
yammesicka: Prefer using constants (there are ready constants in the starlette module)

app/routers/meds.py
```diff
@@ -0,0 +1,424 @@
...
+    datetimes = get_reminder_datetimes(form)
+    for event_time in datetimes:
+        data = {
+            'title': title,
+            'start': event_time,
+            'end': event_time + timedelta(minutes=5),
+            'content': note,
+            'owner_id': user_id,
+        }
+        events.append(create_model(session, Event, **data))
```
yammesicka: Great practice, the unpacking here makes it look great.

app/routers/meds.py
```diff
@@ -0,0 +1,424 @@
...
+def create_events(session: Session, user_id: int,
+                  form: Dict[str, str]) -> None:
+    """Creates reminder events in the DB based on form data.
+
+    Args:
+        session (Session): DB session.
+        user_id (int): ID of user to create events for.
+        form (dict(str, str)): Medication form containing all relevant data.
+    """
+    events = []
```
yammesicka: Are we really need this list?

app/routers/meds.py
```diff
@@ -0,0 +1,424 @@
...
+            reminder = datetime.combine(start.date(), time_obj)
+            reminder = adjust_day(reminder, early, time_obj)
+            if reminder > start:
+                if not validate_first_day_reminder(datetime_obj, time_obj,
+                                                   minimum, maximum):
+                    reminder = get_different_time_reminder(
+                        datetime_obj, minimum, early, late)
+                yield reminder
+                datetime_obj = reminder
+                i += 1
```
yammesicka: This block should be in a separate function.
It would allow you to put guards and create less indented, more fluent and readable code :)

yammesicka (changes requested):
Great job!
What's left is mostly to refactor some styling issues, but it looks like a great work with lots of effort and thought behind it.

yammesicka (approved):

Issue #67, opened 2021-01-17
Support filter events by categories


ivarshav:
This feature will allow to filter an event by category in js.

This feature includes FE changes only.

Blocked by #29 issue.

ivarshav:
Related PR:
https://github.com/PythonFreeCourse/calendar/pull/298

The issue was completed.

PR #298, opened 2021-02-15
Support Filter by Categories

ivarshav:
Allow filtering events in agenda view by categories.
Right now, only filter by category id is available.
In next fase, filter by category name will be supported.

https://github.com/PythonFreeCourse/calendar/issues/67

Demo:
image


app/static/js/categories_filter.js
```diff
@@ -0,0 +1,31 @@
...
+        if (filteredValues.length > 0) {
+            innerResult = '<div class="p-3">' + key + '</div>';
+            for (event of filteredValues) {
+                innerResult += '<div class="event_line" style="background-color:' + event.color + '">';
+                let availability = event.availability ? 'Busy' : 'Free';
+                innerResult += '<div class=' + availability.toLocaleLowerCase() + ' title=' + availability + '></div>';
+                innerResult += '<div><b>' + event.start + ' - <a href="/event/view/' + event.id + '">' + encodeHTML(event.title) + '</a></b><br>';
+                innerResult += '<span class="duration">duration:' + event.duration + '</span>';
+                innerResult += '</div>';
+                innerResult += '</div>';
```
yammesicka: Maybe it's better to use JS `Template ${strings}`?

app/static/js/categories_filter.js
```diff
@@ -0,0 +1,31 @@
+function filterByCategory(events) {
```
yammesicka: Whoa! JS Ninja

app/templates/agenda.html
```diff
@@ -3,32 +3,47 @@
...
 
-  <div class="exact_date pt-3">
-    <div class="mx-3">
-      <a class="btn btn-light btn-outline-primary btn-sm" href="/agenda?days=0#dates">{{ gettext("Today") }}</a>
+    <div class="category_filter">
+      <label for="category" class="sr-only">Filter events by category</label>
+      <input type="text" id="category" name="category" class="form-control-sm"
+             placeholder="category ID" value="{{ category }}"
+             onfocus="this.value=''" required>
+      <button class="btn btn-outline-info btn-sm" onclick='filterByCategory({{ events | tojson | safe }})'>
```
yammesicka: Prefer using JS's `addEventListener` to do this (we do try to separate the view from the logic)

yammesicka (changes requested):
Very nice job! I've added few insights :)

app/static/js/categories_filter.js
ivarshav: Thanks :)

app/static/agenda_style.css
```diff
@@ -6,6 +6,23 @@
...
+  grid-area: header;
+}
+
+.agenda_grid {
+   display: grid;
+   grid-template-areas:
+      "header filter"
+      "sidebar sidebar";
+   grid-template-columns: 5fr 2fr;
+   grid-gap: 0 20px;
```
yammesicka: Prefer `em`s

app/static/js/categories_filter.js
```diff
@@ -0,0 +1,39 @@
...
+    ` : `
+    .event_line {
+      display: none;
+      visibility: hidden;
+    }
+    .event_line[data-name*="${category}" i] {
+      display: grid;
+      visibility: visible;
+    }
+    `;
```
yammesicka: It's usually considered a bad practice to directly edit the CSS.
Maybe you can toggle a class, or a data attribute, and use `.event-line[data-active="on"]` and `.event-line[data-active="off"]`?

app/static/js/categories_filter.js
```diff
@@ -0,0 +1,39 @@
...
+    window.cssFilter.innerHTML = css;
+
+    const allEvents = document.getElementsByClassName("event_line");
+    for (event of allEvents) {
+      event.parentNode.style.display = "none" ;
```
yammesicka: Maybe replace with CSS and dataset?

app/routers/agenda.py
```diff
@@ -47,10 +48,14 @@ def agenda(
...
-            event_obj.start, event_obj.end
-        )
-        events[event_obj.start.date()].append((event_obj, event_duration))
+        event_duration = agenda_events.get_time_delta_string(event_obj.start,
+                                                             event_obj.end)
+        json_event_data = jsonable_encoder(event_obj)
+        json_event_data['duration'] = event_duration
+        json_event_data['start'] = event_obj.start.time().strftime(
+            "%H:%M")
+        events[event_obj.start.date().strftime("%d/%m/%Y")].append(
```
yammesicka: I would separate `event_obj.start.date().strftime("%d/%m/%Y")` to a new line and give it a name, to simplify the reading of this function :)

yammesicka (changes requested):
Great job! I've added few additional notes, mainly about the dynamic edit of the CSS

app/static/js/categories_filter.js
ivarshav: Not sure what do you mean here?

app/static/js/categories_filter.js
ivarshav: I can change the CSS part and instead change in js (something similar to the loops below...), WDYT?

yammesicka: Usually we want the JavaScript to change the value of the `data-` attributes in the HTML, or toggle the classes of the elements in the DOM.
It's better to avoid dynamically editing the CSS using JavaScript, if possible.

Is there a way to move this section of code to the CSS, and to change only the `data-` attributes/classes of the relevant elements?

app/static/js/categories_filter.js
yammesicka: Same as above

ivarshav: Here I change the parent of "event_line" element, which is "wrapper" element. 
This is done in order to hide titles of dates when all events in this day are filtered out.

I'm not sure it can be done in the same way of dataset, as the indicator to know if we should filter is not available for "wrpper" element.
The indicator is the category_id, which we know only in the scope of the event itself and not in the scope of the parent. 

The indicator to hide a "wrpper" element is if all "event_line" elemmts inside are hidden.

ivarshav: Changed as requested, done.

yammesicka (approved):

PR #302, opened 2021-02-16
refactor/jinja-global-variable

issac211:
user global variable

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -36,8 +36,8 @@ set_ui_language()
 
 from app.routers import (  # noqa: E402
     agenda, calendar, categories, celebrity, currency, dayview,
-    email, event, four_o_four, invitation, profile, search,
-    weekview, telegram, whatsapp,
+    email, event, four_o_four, global_variable, invitation, profile,
+    search, weekview, telegram, whatsapp
 )
 
 json_data_loader.load_to_db(next(get_db()))
@@ -53,6 +53,7 @@ routers_to_include = [
     email.router,
     event.router,
     four_o_four.router,
+    global_variable.router,
     invitation.router,
     profile.router,
     salary.router,
@@ -60,6 +61,12 @@ routers_to_include = [
     telegram.router,
     whatsapp.router,
 ]
+
+user_id = 1
+db_gen = get_db()
+session = next(db_gen)
+user = session.query(models.User).filter_by(id=user_id).first()
+templates.env.globals['user'] = user
 
 for router in routers_to_include:
     app.include_router(router)
--- a/app/routers/global_variable.py
+++ b/app/routers/global_variable.py
@@ -1,0 +1,33 @@
+from fastapi import APIRouter, Depends, Request
+from sqlalchemy.orm import Session
+
+from app.database.models import User
+from app.dependencies import get_db, templates
+
+
+router = APIRouter(
+    prefix="/global-variable",
+    tags=["global-variable"],
+    responses={404: {"description": "Not found"}},
+)
+
+
+@router.get("/")
+async def global_var(request: Request, db: Session = Depends(get_db)):
+    user = User(
+        username='test_user',
+        email='my@email.po',
+        password='1a2s3d4f5g6',
+        full_name='My Name',
+        language_id=1,
+        telegram_id='',
+    )
+    user_db = db.query(User).filter_by(username=user.username).first()
+    if not user_db:
+        db.add(user)
+        db.commit()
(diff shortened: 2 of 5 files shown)


app/routers/global_variable.py
```diff
@@ -0,0 +1,33 @@
+from fastapi import APIRouter, Depends, Request
```
PureDreamer: This shouldn't be here, and should be under demos(create it) because it is not supposed to be on the main app

app/templates/base.html
```diff
@@ -54,6 +54,9 @@
                         <li class="nav-item">
                             <a class="nav-link" href="{{ url_for('search') }}">Search</a>
                         </li>
+                        <li class="nav-item">
+                            <a class="nav-link fw-bolder" href="https://forums.pythonic.guru/">The Forum</a>
+                        </li>
```
PureDreamer: Although it's nice to have a link to the forum, this app is not intended for us only.

PureDreamer (commented):
The idea seem very good, yet I can't see how does it work, could you add documentation?
And if it does work the way you wanted, could you add support for showing log-out only for logged-in users, and login/sign up for not logged-in users?

app/main.py
```diff
@@ -45,6 +45,7 @@ def create_tables(engine, psql_environment):
     google_connect, invitation, joke, login, logout, profile,
     register, search, telegram, user, weekview, weight, whatsapp,
 )
+from app.demos import global_variable  # noqa: E402
```
yammesicka: abc

yammesicka (commented):
I find it a bit hard to understand what is the purpose of the feature.
Can you please explain it in a nutshell?

issac211:
> I find it a bit hard to understand what is the purpose of the feature.
> Can you please explain it in a nutshell?

Upload the connected user to a global variable
So that all routers can access it without placing it in their variables.
I therefore place the global variable outside the function of the router.

The router I added is for illustration purposes, so it's inside a demo folder.

ivarshav (approved):

issac211:
https://github.com/PythonFreeCourse/calendar/pull/302#pullrequestreview-594867167
I hope it's clearer now,

in the tests\global_var_testing_routes.py
you can see how the function 'set_global_user_var' can be used in the router.

I changed the code, and used the security module functions to make it more usable in the project.
wen using the 'set_global_user_var' function the template will get the username if the user is logged in

So that this function can be used in routers that still want to identify if the user is connected but do not want to be redirected to the login page if not (This is how it was defined in the security module functions)

app/templates/global_var_test.html
```diff
@@ -0,0 +1,52 @@
+<!DOCTYPE html>
```
yammesicka, line 1: There shouldn't be a file used just for testing in the `templates` directory.
Check the documentation for functions that allow you to load templates from strings instead.

tests/global_var_testing_routes.py
```diff
@@ -0,0 +1,21 @@
+from sqlalchemy.orm import Session
```
yammesicka, line 1: I believe this file can also be deleted

app/internal/global_variable.py
```diff
@@ -0,0 +1,18 @@
+from fastapi import Request
+
+from app.dependencies import templates
+from app.internal.security.ouath2 import (
+    Session, get_jwt_token, get_authorization_cookie
+)
+
+
+async def get_user_for_global_var(db: Session, jwt: str) -> str:
+    jwt_payload = await get_jwt_token(db, jwt)
```
yammesicka, line 10: This will make the site query the database with every request to the site.
It's nullify the advantages of JWT

app/internal/global_variable.py
```diff
@@ -0,0 +1,18 @@
+from fastapi import Request
+
+from app.dependencies import templates
+from app.internal.security.ouath2 import (
+    Session, get_jwt_token, get_authorization_cookie
```
yammesicka, line 5: Use git precommit hooks to order that right and add a trailing comma

yammesicka (changes requested):

app/internal/global_variable.py
issac211: So maybe I could pull the token out of the cookie, and if the token is in the right length (and maybe more tests), I wiil set a global var: 'is_token'.
To verify that there is a logged in user. 
Tell me what you think

yammesicka: Great! You can also take the user details from the JWT, not from the database.

PR #317, opened 2021-02-17
Bugfix/telegram testing

PureDreamer:


tests/asyncio_fixture.py
```diff
@@ -38,6 +38,7 @@ def fake_user_events(session):
         content='test event',
         owner_id=user.id,
         location="Here",
+        is_google_event=False,
```
ivarshav, line 41: Maybe also add it as default value to `create_event`?

PureDreamer: I did a bit more :D

yammesicka (approved):

PR #328, opened 2021-02-20
Telegram v2

leddcode:


tests/test_a_telegram_asyncio.py
```diff
@@ -140,7 +140,7 @@ async def test_user_not_registered(telegram_client):
         assert response.status_code == status.HTTP_200_OK
```
subscorp, line 140: prefer response.ok

tests/test_a_telegram_asyncio.py
```diff
@@ -140,7 +140,7 @@ async def test_user_not_registered(telegram_client):
...
         assert b'Hello, Moshe!' in response.content
         assert b'To use PyLendar Bot you have to register' \
```
subscorp, line 142: I think we shouldn't use \ for line breaks.

subscorp (changes requested):

PR #329, opened 2021-02-20
fix: event backend and frontend creation

imimouni:
fix overrided merges-
location and vc separation
invited email

fix invited emails error when empty

fix vc_link references

fix models.py - make vc link nullable

### DIFF
--- a/app/internal/event.py
+++ b/app/internal/event.py
@@ -9,46 +9,56 @@ from starlette.status import HTTP_400_BAD_REQUEST
 
 from app.database.models import Event
 
-ZOOM_REGEX = re.compile(r'https://.*?\.zoom.us/[a-z]/.[^.,\b\s]+')
+ZOOM_REGEX = re.compile(r"https://.*?\.zoom.us/[a-z]/.[^.,\b\s]+")
 
 
 def raise_if_zoom_link_invalid(vc_link):
     if ZOOM_REGEX.search(vc_link) is None:
-        raise HTTPException(status_code=HTTP_400_BAD_REQUEST,
-                            detail="VC type with no valid zoom link")
+        raise HTTPException(
+            status_code=HTTP_400_BAD_REQUEST,
+            detail="VC type with no valid zoom link",
+        )
 
 
 def get_invited_emails(invited_from_form: str) -> List[str]:
     invited_emails = []
-    for invited_email in invited_from_form.split(','):
-        invited_email = invited_email.strip()
-        try:
-            validate_email(invited_email, check_deliverability=False)
-        except EmailSyntaxError:
-            logging.exception(f'{invited_email} is not a valid email address')
-            continue
-        invited_emails.append(invited_email)
+    if not invited_from_form:
+        invited_emails.append("")
+    else:
+        for invited_email in invited_from_form.split(","):
+            invited_email = invited_email.strip()
+            try:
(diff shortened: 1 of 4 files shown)


app/internal/event.py
```diff
@@ -9,46 +9,56 @@
...
+    else:
+        for invited_email in invited_from_form.split(","):
+            invited_email = invited_email.strip()
+            try:
+                validate_email(invited_email, check_deliverability=False)
+            except EmailSyntaxError:
+                logging.exception(
+                    f"{invited_email} is not a valid email address",
+                )
+                continue
```
subscorp: Probably shouldn't use continue or break.

subscorp (changes requested):
Looks good! :)

app/templates/partials/calendar/event/edit_event_details_tab.html
```diff
@@ -40,25 +40,22 @@
...
-  <input type="text" name="location" placeholder="VC URL/Location">
+    <label for="location">Location </label>
+    <input type="text" id="location" name="location" placeholder="Location">
 </div>
 
-<div class="form_row textarea">
-  <textarea id="say" name="description" placeholder="Description"></textarea>
+<div class="form_row">
+    <label for="vc_link">VC link: </label>
+    <input type="text" id="vc_link" name="vc_link" placeholder="VC URL">
```
subscorp, line 49: Use dashes in HTML ids: vc-link instead of vc_link.

subscorp (changes requested):

app/internal/event.py
imimouni: Not my code.
I added the if  not invited from form as this was producing an error for empty "invitees" field

imimouni: (so I am not sure what the effect will be, need to see who wrote this originally)

app/internal/event.py
```diff
@@ -9,46 +9,56 @@
...
-    for invited_email in invited_from_form.split(','):
-        invited_email = invited_email.strip()
-        try:
-            validate_email(invited_email, check_deliverability=False)
-        except EmailSyntaxError:
-            logging.exception(f'{invited_email} is not a valid email address')
-            continue
-        invited_emails.append(invited_email)
+    if not invited_from_form:
+        invited_emails.append("")
```
yammesicka: `return [""]` instead, and unindent the `else` clause

yammesicka (commented):
Great work! Please just use guard instead of the long if/else in `get_invited_emails`

app/internal/event.py
imimouni: changed this a bit. to guard clause as Yam asked + instead of continue used - try, except, else

app/templates/partials/calendar/event/edit_event_details_tab.html
imimouni: This is really problematic.
I've looked this through and though you are right this is a problem it exists all over the project in different tags and affects many many files and not related to the purpose of this PR (fixing create event).
Maybe it's necessary to open a new PR to fix this throughout the project . @yammesicka What do you think?

app/internal/event.py
imimouni: Done :) (I used try, except else inside the for loop)

yammesicka (approved):

PR #354, opened 2021-02-28
Feat/jinja global variable

kobyfogel:
logged in user as a jinja global variable


app/main.py
```diff
@@ -6,21 +6,22 @@
 from fastapi.staticfiles import StaticFiles
 from sqlalchemy.orm import Session
 
+import app.internal.features as internal_features
```
yammesicka, line 9: Move to the correct place

yammesicka (approved):

